PR #15, opened 2021-01-19
Renamed isLockTx method to isPegInTx

Vovchyk:
*rskj: active-block-reg*

### DIFF
--- a/src/main/java/co/rsk/federate/BtcToRskClient.java
+++ b/src/main/java/co/rsk/federate/BtcToRskClient.java
@@ -620,7 +620,7 @@ public class BtcToRskClient implements BlockListener, TransactionListener {
     private boolean isTxProcessable(BtcTransaction btcTx, TxSenderAddressType txSenderAddressType) {
         // If the tx is a release it means we are receiving change (or migrating funds)
         // If the tx is a release it should be processable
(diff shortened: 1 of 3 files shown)


Vovchyk:
pipeline: run

Vovchyk:
pipeline: run

marcos-iov (approved):

Vovchyk:
pipeline: run

marcos-iov (dismissed):

.circleci/config.yml
```diff
@@ -27,6 +27,8 @@ jobs:
           name: Clone rskj repo
           command: |
             git clone https://github.com/rsksmart/rskj.git
+            cd rskj
+            git checkout ddf2e391eeaaef1575cfc64d70ab065be73fa6fb
```
josedahlquist, line 31: this should be removed
```suggestion
```
circle was updated to search for a matching rskj branch to perform the build

Vovchyk: removed this part and added `rskj: active-block-reg` in the desc. Looks like it doesn't pick the rskj branch up.

marcos-iov: You should rename this branch to `active-block-reg` since CircleCI searches for a branch in rskj with a name equal to the branch it's running in powpeg. If it can't find it then it builds against `master`

Keep the comment `rskj: active-block-reg` in the description, that's for the pipeline

PR #18, opened 2021-01-26
Reproducible action

bcodesido:
This PR adds a Github action that when a release is created executes the reproducible build.


.github/workflows/reproducible.yml
```diff
@@ -0,0 +1,46 @@
...
+  release:
+    types:
+      - created
+
+jobs: 
+  build:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Chekcout
+        uses: actions/checkout@v2
```
lucasvuotto, line 12: Why are we doing a checkout if we clone afterwards?

.github/workflows/reproducible.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: sudo apt-get update -y && sudo apt-get install -y git curl gnupg
+
+      - name: Download Release key
+        run: gpg --keyserver https://secchannel.rsk.co/release.asc --recv-keys 1A92D8942171AFA951A857365DECF4415E3B8FA4
+
+      - name: Check fingerprint
+        run: gpg --finger 1A92D8942171AFA951A857365DECF4415E3B8FA4
+
+      - name: Verify SHA256SUM
+        run: gpg --verify SHA256SUMS.asc && sha256sum --check SHA256SUMS.asc
```
lucasvuotto, line 28: This will spam the output with malformed lines. It would be best invoked as
```
gpg --verify --output SHA256SUMS SHA256SUMS.asc && sha256sum --check SHA256SUMS
```

.github/workflows/reproducible.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Build powpeg-node
+        run: ./configure.sh && ./gradlew --no-daemon clean build -x test
+
+      - name: Calculate hashes
+        working-directory: build/libs
+        run: sha256sum * > hashes
+
+      - name: Print hashes
+        run: echo $(cat build/libs/hashes)
```
lucasvuotto, line 38: `sha256sums * | tee hashes` will achieve the same effect in just one step. Also, `echo $(cat file)` is a bad idiom for `cat file`.

.github/workflows/reproducible.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: ${{ steps.vars.outputs.tag }}
+          path: build/libs/
+
+      - name: Use openjdk container version 335627a2118556b3f412287e08ac7febb8ecc46325dc6b3570db56f32d33938f
+        uses: docker://openjdk@sha256:335627a2118556b3f412287e08ac7febb8ecc46325dc6b3570db56f32d33938f
```
lucasvuotto, line 46: According to https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses, you're instruction this step to run in that Docker container, instead of the whole action execution.

.github/workflows/reproducible.yml
```diff
@@ -0,0 +1,46 @@
+name: "Reproducible build"
+on:
+  release:
+    types:
+      - created
```
lucasvuotto, line 5: I have some doubts with this. According to https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#release, `created` means

> a draft is saved, or a release or pre-release is published without previously being saved as a draft

I think it isn't a common workflow for us, but will it trigger if we decide to publish a draft?

bcodesido: Yes, already trigger it

.github/workflows/reproducible.yml
bcodesido: fixed

PR #96, opened 2022-09-16
Adds 'activations' in BridgeEventLoggerImpl class used in BtcReleaseC…

jeremy-then:
Adds 'activations' in BridgeEventLoggerImpl class used in BtcReleaseClientStorageSynchronizerTest.java file


marcos-iov:
Could please change it to a draft PR? So we don't accidentally merge it, just in case

PR #99, opened 2022-09-26
Implement new difficulty cap setting when accumulating difficulty

kelvinator07:


nathanieliov:
@kelvinator07 the build by CircleCI is failing. If the reason why this can not be built is related to the need of using a specific branch from RSKJ, instead of `master`, please create a branch with the same name as this branch(`implement-difficulty-cap`), so it uses that rskj branch with your latest changes to compile.

jeremy-then (approved):
LGTM.

marcos-iov (approved):

PR #159, opened 2023-04-21
Update bridge refactor

marcos-iov:
Fix so that if a transaction fails to be registered in the Bridge the PowPeg will continue trying to register the remaining transactions.

### DIFF
--- a/src/main/java/co/rsk/federate/BtcToRskClient.java
+++ b/src/main/java/co/rsk/federate/BtcToRskClient.java
@@ -23,7 +23,6 @@ import org.bitcoinj.core.*;
 import org.bitcoinj.store.BlockStoreException;
 import org.ethereum.config.blockchain.upgrades.ActivationConfig;
 import org.ethereum.config.blockchain.upgrades.ConsensusRule;
-import org.ethereum.core.Blockchain;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


src/main/java/co/rsk/federate/BtcToRskClient.java
```diff
@@ -426,9 +467,11 @@ private StoredBlock findBridgeBtcBlockchainMatchingAncestorUsingBlockLocator() t
         return matchedBlock;
     }
 
-    public void updateBridgeBtcTransactions() throws BlockStoreException {
+    protected void updateBridgeBtcTransactions() throws BlockStoreException {
```
nathanieliov, line 470:
```suggestion
    protected void updateBridgeBtcTransactions() {
```

src/main/java/co/rsk/federate/BtcToRskClient.java
```diff
@@ -442,143 +485,184 @@ public void updateBridgeBtcTransactions() throws BlockStoreException {
...
+                            bestChainHeight
+                        );
+                    }
                 }
+            } catch (Exception e) {
+                logger.error(
+                    "[updateBridgeBtcTransactions] An error occurred while informing transaction {} to the Bridge. {}",
+                    txHash,
+                    e.getMessage()
+                );
```
nathanieliov, line 624: What do you think about moving this to a separate method and making it there easier to follow? For me, it's hard to follow those big block of code,I mean the `try` block.

kelvinator07: I Agree With Nath, We Could Move Some Of This Logic To Separate Methods, It Can Be Handled Now Or We Create Tasks To Handle It Later

kelvinator07:
I'd Also Suggest We Handle The Multiple Try's And Catches In `updateBridge` In The Callees Methods Instead Of Inside The `updateBridge`

src/main/java/co/rsk/federate/BtcToRskClient.java
marcos-iov: Thanks for the comments! I'll create a new story to handle this later

nathanieliov (approved):

PR #177, opened 2023-06-16
Reduce hsm calls

kelvinator07:


src/main/java/co/rsk/federate/signing/hsm/advanceblockchain/HSMBookkeepingService.java
```diff
@@ -134,11 +135,13 @@ protected void informConfirmedBlockHeaders() {
         informing = true;
         logger.info("Starting HSM bookkeeping process");
         try {
-            Block hsmCurrentBestBlock = getHsmBestBlock();
             if (hsmCurrentBestBlock == null) {
-                logger.error("[informConfirmedBlockHeaders] Can't found HSM Best Block in blockStore.");
-                informing = false;
-                return;
+                hsmCurrentBestBlock = getHsmBestBlock();
+                if (hsmCurrentBestBlock == null) {
```
marcos-iov, line 140: Under what circumstance would the HSM best block not be found in the block store?

That seems like a ver abnormal situation, perhaps we could do this validation inside `getHsmBestBlock` method and if not found then throw an exception

kelvinator07: Updated

src/main/java/co/rsk/federate/signing/hsm/advanceblockchain/HSMBookkeepingService.java
```diff
@@ -5,6 +5,7 @@
 import co.rsk.federate.signing.hsm.client.HSMBookkeepingClient;
 import co.rsk.federate.signing.hsm.message.AdvanceBlockchainMessage;
 import co.rsk.net.NodeBlockProcessor;
+import org.bitcoinj.store.BlockStoreException;
```
marcos-iov, line 8: Not sure this is the right exception to throw here. This is a bitcoinj exception, here we are dealing with RSK blocks and RSK storage

src/main/java/co/rsk/federate/signing/hsm/advanceblockchain/HSMBookkeepingService.java
```diff
@@ -113,9 +115,15 @@ public boolean isStarted() {
         return started;
     }
 
-    private Block getHsmBestBlock() throws HSMClientException {
+    private Block getHsmBestBlock() throws HSMClientException, BlockStoreException {
```
marcos-iov, line 118: What happens with the BlockStoreException? Is it caught at any point or does it bring down the node?

src/test/java/co/rsk/federate/signing/hsm/advanceblockchain/HSMBookkeepingServiceTest.java
```diff
@@ -9,6 +9,7 @@
 import co.rsk.federate.signing.hsm.message.PowHSMState;
 import co.rsk.federate.signing.utils.TestUtils;
 import co.rsk.net.NodeBlockProcessor;
+import org.bitcoinj.store.BlockStoreException;
```
marcos-iov, line 12: unused import?

src/main/java/co/rsk/federate/signing/hsm/advanceblockchain/HSMBookkeepingService.java
kelvinator07: Yes, Not The Right Exception. Updated

src/test/java/co/rsk/federate/signing/hsm/advanceblockchain/HSMBookkeepingServiceTest.java
kelvinator07: Removed

src/main/java/co/rsk/federate/signing/hsm/advanceblockchain/HSMBookkeepingService.java
kelvinator07: Yes It Caught [Here](https://github.com/rsksmart/powpeg-node/blob/b0c89b2840d8fb33e822f7a0b9ef63771d474628/src/main/java/co/rsk/federate/signing/hsm/advanceblockchain/HSMBookkeepingService.java#L175), And Doesnt Bring Down The Node

marcos-iov: I see, thanks

src/main/java/co/rsk/federate/signing/hsm/BlockStoreException.java
```diff
@@ -0,0 +1,12 @@
+package co.rsk.federate.signing.hsm;
```
marcos-iov, line 1: I'm not sure about this. The BlockStore is part of RSK, and we are adding here a BlockStoreException inside the HSM signing package. Doesn't seem related.

Perhaps we could use the already existing HSMBlockchainBookkeepingRelatedException, that one seems to make sense for this case. What do you think?

kelvinator07: Yes I Think Using The Existing HSMBlockchainBookkeepingRelatedException Is Better. Updated

marcos-iov (approved):

PR #197, opened 2023-08-21
Refactor hsm commands to enum

kelvinator07:


src/main/java/co/rsk/federate/signing/HSMCommand.java
```diff
@@ -0,0 +1,25 @@
+package co.rsk.federate.signing;
+
+/**
+ * Created by Kelvin Isievwore on 21/08/2023.
+ */
+public enum HSMCommand {
```
marcos-iov, line 6: Should it be protected?

marcos-iov (approved):
Well done! Looks really good

src/main/java/co/rsk/federate/signing/HSMCommand.java
kelvinator07: Enums Can Only Be Public

src/main/java/co/rsk/federate/signing/hsm/client/HSMClientProtocol.java
```diff
@@ -100,35 +99,35 @@ public ObjectNode buildCommand(String commandName, int version) {
...
+                        throw (JsonRpcException) cause;
                     }
                     if (cause instanceof HSMDeviceNotReadyException) {
-                        throw (HSMDeviceNotReadyException)cause;
+                        throw (HSMDeviceNotReadyException) cause;
                     }
                     if (cause instanceof HSMClientException) {
-                        throw (HSMClientException)cause;
+                        throw (HSMClientException) cause;
                     }
```
nathanieliov, line 123: What if the cause isn't any of those exception types? `result` method variable will be null, and the code will continue until the end. Why not add a default when the cause is none of those types?

src/main/java/co/rsk/federate/signing/hsm/client/HSMSigningClientV1.java
```diff
@@ -66,14 +69,14 @@ public byte[] getPublicKey(String keyId) throws HSMClientException {
 
     @Override
     public HSMSignature sign(String keyId, SignerMessage message) throws HSMClientException {
-        byte[] messageBytes = ((SignerMessageV1)message).getBytes();
+        byte[] messageBytes = ((SignerMessageV1) message).getBytes();
```
nathanieliov, line 72:
```suggestion
        byte[] messageBytes = message.getBytes();
```

nathanieliov (commented):
Well done. Left two comments.

src/main/java/co/rsk/federate/signing/hsm/client/HSMClientProtocol.java
kelvinator07: This Is Out Of The Scope For This Task, However, There's An External Try Catch To Handle It

src/main/java/co/rsk/federate/signing/hsm/client/HSMSigningClientV1.java
kelvinator07: Updated

src/main/java/co/rsk/federate/signing/hsm/client/HSMClientProtocol.java
nathanieliov: The external try catch will handle it **only if** the exception is not an `ExecutionException,` otherwise the flow will continue and call `validateResponse` passing a null.

nathanieliov: I am closing this discussion since we agreed to handle this in another ticket.

nathanieliov (approved):
LGTM

PR #211, opened 2023-10-06
Use legacy methods from PegUtilsLegacy class

nathanieliov:
*rskj:move-legacy-methods-to-pegUtils*


marcos-iov:
Would it make sense that instead of calling the individual PegUtilsLegacy methods (isPegin, isPegout) we simply called getTransactionType?

nathanieliov:
> Would it make sense that instead of calling the individual PegUtilsLegacy methods (isPegin, isPegout) we simply called getTransactionType?

@marcos-iov, that would be ideal. But that's not possible since the legacy getTransactionType depends on bridgeStorageProvider, which powpeg node doesn't have access to

nathanieliov:
Closing this PR due to is duplicated. Let's continue discussions in #212

nathanieliov:
Closing this PR due to is duplicated. Let's continue discussions at https://github.com/rsksmart/powpeg-node/pull/212

PR #226, opened 2023-12-14
Create method in BtcReleaseClient to extract default redeem script instead of using instanceof

julia-zack:


src/main/java/co/rsk/federate/btcreleaseclient/BtcReleaseClient.java
```diff
@@ -451,6 +451,13 @@ protected Script extractStandardRedeemScript(Script redeemScript) {
         return parser.extractStandardRedeemScript();
     }
 
+    protected Script extractDefaultRedeemScript(Federation federation) {
```
marcos-iov, line 454:
```suggestion
    private Script extractDefaultRedeemScript(Federation federation) {
```

julia-zack: done

nathanieliov (approved):

PR #229, opened 2024-01-10
Adapt to federation args creation

julia-zack:
In rskj project, FederationArgs // ErpFederationArgs objects were created to reduce the amount of params we pass to the Federation // ErpFederation constructors.
This pr adapts to that change

### DIFF
--- a/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
+++ b/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
@@ -20,10 +20,10 @@ package co.rsk.federate;
 
 import co.rsk.bitcoinj.core.Address;
 import co.rsk.bitcoinj.core.BtcECKey;
+import co.rsk.bitcoinj.core.NetworkParameters;
 import co.rsk.config.BridgeConstants;
-import co.rsk.peg.federation.Federation;
-import co.rsk.peg.federation.FederationFactory;
-import co.rsk.peg.federation.FederationMember;
+import co.rsk.peg.federation.*;
+import org.ethereum.config.blockchain.upgrades.ActivationConfig;
 import org.ethereum.crypto.ECKey;
 
 import java.time.Instant;
@@ -74,13 +74,10 @@ public class FederationProviderFromFederatorSupport implements FederationProvide
         }
         Instant creationTime = federatorSupport.getFederationCreationTime();
         long creationBlockNumber = federatorSupport.getFederationCreationBlockNumber();
+        NetworkParameters btcParams = federatorSupport.getBtcParams();
+        FederationArgs federationArgs = new FederationArgs(members, creationTime, creationBlockNumber, btcParams);
 
-        Federation initialFederation = FederationFactory.buildStandardMultiSigFederation(
-            members,
-            creationTime,
-            creationBlockNumber,
-            federatorSupport.getBtcParams()
-        );
+        Federation initialFederation = FederationFactory.buildStandardMultiSigFederation(federationArgs);
 
         return getExpectedFederation(initialFederation, getActiveFederationAddress());
     }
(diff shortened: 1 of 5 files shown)


marcos-iov (approved):

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -156,31 +150,26 @@ private Federation getExpectedFederation(Federation initialFederation, Address e
...
+        long creationBlockNumber = initialFederation.getCreationBlockNumber();
+        NetworkParameters btcParams = federatorSupport.getBtcParams();
+        List<BtcECKey> erpPubKeys = bridgeConstants.getErpFedPubKeysList();
+        long activationDelay = bridgeConstants.getErpFedActivationDelay();
+        ActivationConfig.ForBlock activations = federatorSupport.getConfigForBestBlock();
+
+        ErpFederationArgs erpFederationArgs =
+            new ErpFederationArgs(members, creationTime, creationBlockNumber, btcParams, erpPubKeys, activationDelay);
+
+        // If addresses match build a Non-Standard ERP federation
```
jeremy-then, line 164: Following the clean code advice we have seen recently, I don't think we need these types of comments. The code below is easy to read and says clearly what it does:

```java
if (nonStandardErpFederation.getAddress().equals(expectedFederationAddress)) {
            return nonStandardErpFederation;
}
```

I think this is clear enough, not need for a documentation comment in this case.

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -156,31 +150,26 @@ private Federation getExpectedFederation(Federation initialFederation, Address e
...
+        ErpFederation nonStandardErpFederation = FederationFactory.buildNonStandardErpFederation(erpFederationArgs, activations);
+
+        if (nonStandardErpFederation.getAddress().equals(expectedFederationAddress)) {
+            return nonStandardErpFederation;
         }
 
         // Finally, try building a P2SH ERP federation
```
jeremy-then, line 171: The same here.

src/test/java/co/rsk/federate/FederationProviderFromFederatorSupportTest.java
```diff
@@ -121,7 +121,7 @@ void getActiveFederation_erp_federation() {
         ActivationConfig.ForBlock configMock = mock(ActivationConfig.ForBlock.class);
         when(configMock.isActive(RSKIP123)).thenReturn(true);
 
-        Federation expectedFederation = createErpFederation(
+        Federation expectedFederation = createNonStandardErpFederation(
             getFederationMembersFromPks(1, 1000, 2000, 3000, 4000, 5000),
```
jeremy-then, line 125: No big deal, but I think these values for a btc private key look unrealistic. I would suggest having valid btc private keys, probably declared as static constant at the beginning of this class. To keep everything as real as possible.

src/test/java/co/rsk/federate/FederationWatcherTest.java
```diff
@@ -31,26 +32,23 @@
...
+    List<FederationMember> federation1Members = getFederationMembersFromPksForBtc(1000, 2000, 3000, 4000);
+    Instant federation1CreationTime = Instant.ofEpochMilli(5005L);
+    long creationBlockNumber = 0L;
+    NetworkParameters btcParams = NetworkParameters.fromID(NetworkParameters.ID_REGTEST);
+
+    FederationArgs federation1Args = new FederationArgs(federation1Members, federation1CreationTime, creationBlockNumber, btcParams);
+    private final Federation federation1 = FederationFactory.buildStandardMultiSigFederation(federation1Args);
+
+    List<FederationMember> federation2Members = getFederationMembersFromPksForBtc(2000, 3000, 4000, 5000, 6000, 7000);
+    Instant federation2CreationTime = Instant.ofEpochMilli(15300L);
```
jeremy-then, line 44: Do these values have a meaning? `15300L` is something important or could it be any other random value? If it needs to be exactly this value, then we could put it in a static final variable. If not, that's fine.

jeremy-then (approved):
LGTM. Left a couple of optional suggestions to discuss.

PR #233, opened 2024-02-01
Build and push docker image on tag

ATarrio:
I've added a new github workflow to generate and push a docker image to use the PowPeg-Node in.

The new workflow logs in on both DockerHub and GitHub Container Registry, builds the code, builds the docker image, then pushes it to both repos. For now, the action will be triggered when a tag is pushed.

The required secrets needed to be set up in this repo for this to work are:
DOCKERHUB_USERNAME : username to login on dockerhub
DOCKERHUB_TOKEN : token to use to login on dockerhub
Please contact the DevOps team to request these credentials

The images will be pushed as:
rsksmart/powpeg-node:<RELEASE_NAME>
ghcr.io/rsksmart/powpeg-node:<RELEASE_NAME>

### DIFF
--- a/.github/workflows/build-push-docker.yml
+++ b/.github/workflows/build-push-docker.yml
@@ -5,8 +5,8 @@ on:
       - '*'
 
 env:
-  GHCR_TAG: "ghcr.io/rsksmart/powpeg-node"
-  DOCKERHUB_TAG: "rsk/powpeg-node"
+  GHCR_REPO: "ghcr.io/rsksmart/powpeg-node"
+  DOCKERHUB_REPO: "rsk/powpeg-node"
 
 jobs:
   build:
@@ -24,8 +24,8 @@ jobs:
         uses: docker/metadata-action@v5
         with:
           images: |
-            ${{ env.DOCKERHUB_TAG }}
-            ${{ env.GHCR_TAG }}
+            ${{ env.DOCKERHUB_REPO }}
+            ${{ env.GHCR_REPO }}
           tags: |
             type=ref,event=tag
             type=edge
@@ -60,5 +60,5 @@ jobs:
 
       - name: Push Images
         run: |
-          docker push ${{ env.DOCKERHUB_TAG }}
-          docker push ${{ env.GHCR_TAG }}
+          docker push ${{ env.DOCKERHUB_REPO }}
+          docker push ${{ env.GHCR_REPO }}
--- a/.github/workflows/build-push-docker.yml
+++ b/.github/workflows/build-push-docker.yml
@@ -1,0 +1,64 @@
+name: Docker Image Build and Push
+on:
+  push:
+    tags:
+      - '*'
+
+env:
+  GHCR_TAG: "ghcr.io/rsksmart/powpeg-node"
+  DOCKERHUB_TAG: "rsk/powpeg-node"
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set version
+        id: vars
+        run: echo ::set-output name=tag::${GITHUB_REF#refs/*/}
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ${{ env.DOCKERHUB_TAG }}
+            ${{ env.GHCR_TAG }}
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN useradd -ms /sbin/nologin -d /var/lib/rsk rsk
+USER rsk
+
+WORKDIR /var/lib/rsk
+COPY --from=build --chown=rsk:rsk /home/rsk/rsk.jar ./
+
+ENV DEFAULT_JVM_OPTS="-Xss4M"
+ENV RSKJ_SYS_PROPS="-Dlogback.configurationFile='/etc/rsk/logback.xml' -Drsk.conf.file=/etc/rsk/node.conf"
+ENV RSKJ_CLASS=co.rsk.federate.FederateRunner
+ENV RSKJ_OPTS=""
```
lucasvuotto, line 35: Any reason why not to sync this one with RskJ one?

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN useradd -ms /bin/bash rsk
+USER rsk
+
+WORKDIR /home/rsk
+COPY --chown=rsk:rsk . ./
+
+RUN gpg --keyserver https://secchannel.rsk.co/SUPPORT.asc --recv-keys 1DC9157991323D23FD37BAA7A6DBEAC640C5A14B && \
+    gpg --verify --output SHA256SUMS SHA256SUMS.asc && \
+    sha256sum --check SHA256SUMS && \
+    ./configure.sh && \
```
apancorb, line 19: Not sure if as part of the build command you will also need to add the `DONT-COMMIT-settings.gradle` file in order to build, as in [here](https://github.com/rootstock/fed-dockerized-test/blob/3d83774d6c8b9340fdc8013c256b3b0474c0a10c/powpeg-node/Dockerfile#L47).

Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM openjdk:11-jdk-slim-buster AS build
```
nathanieliov, line 1: Any reason why you are using `11-jdk` instead of `openjdk:8-jdk-slim-buster` like the reproducible builds?

Dockerfile
ATarrio: No idea, I just did it same as it was done in powpeg-node-setup. But I can change it if you think it's worthwile.

Dockerfile
ATarrio: No, I believe I took that from the rskj Dockerfile, but I'll try using version 8 and correct this.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+LABEL org.opencontainers.image.authors="ops@iovlabs.org"
+
+RUN useradd -ms /sbin/nologin -d /var/lib/rsk rsk
+USER rsk
+
+WORKDIR /var/lib/rsk
+COPY --from=build --chown=rsk:rsk /home/rsk/rsk.jar ./
+
+ENV DEFAULT_JVM_OPTS="-Xss4M"
+ENV RSKJ_SYS_PROPS="-Dlogback.configurationFile='/etc/rsk/logback.xml' -Drsk.conf.file=/etc/rsk/node.conf"
```
wilmerrootstock, line 33: Is this the correct path for `logback.xml` and `node.conf` files?

C.C.: @nathanieliov @jeremy-then @marcos-iov

wilmerrootstock (changes requested):

Dockerfile
apancorb: I see that makes sense, thanks

.github/workflows/build-push-docker.yml
```diff
@@ -0,0 +1,64 @@
+name: Docker Image Build and Push
+on:
+  push:
+    tags:
+      - '*'
+
+env:
+  GHCR_TAG: "ghcr.io/rsksmart/powpeg-node"
+  DOCKERHUB_TAG: "rsk/powpeg-node"
```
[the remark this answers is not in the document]
ATarrio: Nitpicking maybe, but correct. Fixed :)

Dockerfile
nathanieliov: Seems correct to me.

nathanieliov:
pipeline:run

ATarrio:
pipeline:run

PR #237, opened 2024-04-02
POC. DO NOT MERGE.

jeremy-then:


src/main/java/co/rsk/federate/FederatorSupport.java
```diff
@@ -151,6 +161,80 @@ public StateForFederator getStateForFederator() {
...
+        Map<String, Utxo> utxoMap = utxos.stream().collect(Collectors.toMap(Utxo::getBtcTxHash, utxo -> utxo));
+
+        return utxoMap;
+
+    }
+
+    private String getUtxoJsonAtBlockRcpCall(long atBlockNumber) {
+        try {
+
+            URL url = new URL("http://localhost:4450");
```
jeremy-then, line 179: As seen here, this example only works for a local node running on port 4450. This is an example to show the POC, no need to do an actual implementation at the moment.

jeremy-then: Also, we could find a better way to do the rpc call.

PR #238, opened 2024-04-03
Refactoring BridgeConstants Classes from RSKj

wilmerrootstock:
- Changing BridgeConstants classes (regtest, testnet, mainnet, devnet) imports from co.rsk.config package to co.rsk.peg package.
- Doing the same inside the test section.

We are passing all the objects to the Peg package to put all objects related to Powpeg together into RSKj. With this, we are taking a step forward to decouple Powpeg.


marcos-iov (approved):

nathanieliov (approved):
LGTM

PR #251, opened 2024-04-26
Fix misleading pegout logs and variables

nathanieliov:
- Update misleading pegout logs with meaningful log messages according to the context.
- Also, update misleading pegout variable names with meaningful names according to the context.


nathanieliov:
pipeline:run

marcos-iov (approved):

PR #257, opened 2024-05-13
Implement cache to prevent Pegnatories signing the same pegout

apancorb:
### Summary

We noticed that in a period of 10 Rootstock blocks there are multiple unnecessary calls to add_signature Bridge method. This stresses the PowHSM devices, reducing their utility life. It also makes the pegnatories waste gas by doing unnecessary transactions.

In this PR, we include the implementation of the cache as well with its interface and factory class following a singleton pattern. Under the hood, the cache will use a `Map<Keccak256, Long>`, where the key represents the pegout creation rootstock transaction hash and the value represents the timestamp when the `addSignature` transaction call was sent to the Bridge.

### Test Plan

Added corresponding unit tests and verified with: `./gradlew test`


src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Retrieves an instance of PegoutSignedCache with the specified Time To Live
+   * (TTL). If an instance does not already exist, it is lazily initialized.
+   * 
+   * @param ttl The duration for which the cache entries are considered valid.
+   * @return An instance of PegoutSignedCache.
+   * @throws IllegalArgumentException If the supplied TTL value is invalid.
+   */
+  public static PegoutSignedCache getInstance(final Duration ttl) throws IllegalArgumentException {
+    if (!isValidTtl(ttl)) {
```
julia-zack, line 27: just as we discussed the last clean-code forum, we should avoid negative conditionals.
I'd suggest you call the method something like `ttlIsNotValid`

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,40 @@
...
+  public void put(Keccak256 pegoutCreationRskTxHash) {
+    Long currentTimestamp = System.currentTimeMillis();
+
+    Optional.ofNullable(pegoutCreationRskTxHash)
+        .ifPresent(rskTxHash -> cache.put(rskTxHash, currentTimestamp));
+  }
+
+  private boolean isValidTimestamp(Long currentTimestamp, Long timestamp) {
+    return currentTimestamp != null && timestamp != null &&
+        currentTimestamp - timestamp <= ttl.toMillis();
```
julia-zack, line 38: i'd convert this line into a variable, something like 
`timestampIsNotExpired = currentTimestamp - timestamp <= ttl.toMillis()`

so is easier to understand

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * Retrieves an instance of PegoutSignedCache with the specified Time To Live
+   * (TTL). If an instance does not already exist, it is lazily initialized.
+   * 
+   * @param ttl The duration for which the cache entries are considered valid.
+   * @return An instance of PegoutSignedCache.
+   * @throws IllegalArgumentException If the supplied TTL value is invalid.
+   */
+  public static PegoutSignedCache getInstance(final Duration ttl) throws IllegalArgumentException {
```
julia-zack, line 26: maybe directly call the variable `timeToLive` ?
So we don't have to check the documentation to understand the name

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
```diff
@@ -0,0 +1,48 @@
...
+    if (!isValidTtl(ttl)) {
+      String message = String.format(
+          "Invalid pegouts signed cache TTL value in minutes supplied: %d",
+          ttl != null ? ttl.toMinutes() : null);
```
julia-zack, line 30: this check could be done in a variable too, `ttlInMinutes = ttl != null ? ttl.toMinutes() : null`

that way it can be reused when logging too :)

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+  void hasAlreadyBeenSigned_shouldReturnFalse_whenPegoutCreationRskTxHashIsNotCached() {
+    boolean result = pegoutSignedCache.hasAlreadyBeenSigned(PEGOUT_CREATION_RSK_HASH);
+
+    assertFalse(result);
+  }
+
+  @Test
+  void hasAlreadyBeenSigned_shouldReturnFalse_whenCacheContainsInvalidTimestamp() {
+    Long currentTimestamp = System.currentTimeMillis();
+    Long timestamp = currentTimestamp - Duration.ofMinutes(60).toMillis();
```
julia-zack, line 54:
```
Long timestampHigherThanTTL = Duration.ofMinutes(60).toMillis()
Long expiredTimestamp = currentTimestamp - timestampHigherThanTTL;
```

wdyt?

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+  @Test
+  void hasAlreadyBeenSigned_shouldReturnFalse_whenPegoutCreationRskTxHashIsNotCached() {
+    boolean result = pegoutSignedCache.hasAlreadyBeenSigned(PEGOUT_CREATION_RSK_HASH);
+
+    assertFalse(result);
+  }
+
+  @Test
+  void hasAlreadyBeenSigned_shouldReturnFalse_whenCacheContainsInvalidTimestamp() {
```
julia-zack, line 52: maybe also naming it `...whenCacheContainsExpiredTimestamp` (instead of `InvalidTimestamp`) is more accurate.
wdyt?

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+  private static final Duration DEFAULT_TTL = Duration.ofMinutes(30);
+  private static final Keccak256 PEGOUT_CREATION_RSK_HASH = TestUtils.createHash(1);
+
+  private final Map<Keccak256, Long> cache = mock(Map.class);
+  private final PegoutSignedCacheImpl pegoutSignedCache = new PegoutSignedCacheImpl(DEFAULT_TTL);
+
+  @BeforeEach
+  void setUp() throws NoSuchFieldException, IllegalAccessException {
+    Field field = pegoutSignedCache.getClass().getDeclaredField("cache");
```
marcos-iov, line 30: Not sure about using this, I think we can set up the different test scenarios by simply calling `put` method. It's more realistic that way

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
julia-zack: very good one

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
```diff
@@ -0,0 +1,48 @@
+package co.rsk.federate.btcreleaseclient.cache;
+
+import co.rsk.panic.PanicProcessor;
```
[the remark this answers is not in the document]
apancorb: Sounds good, I have updated to just use exceptions and log messages.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
apancorb: `ttl` is a common way to identify "time to live" especially in a caching context, lets keep it as is.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
apancorb: Sounds great, thank you both! I just updated the method to reflect what y'all described. My only contribution is to make the private method name `assertValidTtl`. Let me know what you think.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
apancorb: With the new changes I don't think this applies anymore, wdyt?

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+import co.rsk.crypto.Keccak256;
+import java.time.Duration;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+class PegoutSignedCacheImpl implements PegoutSignedCache {
+
+  private final Map<Keccak256, Long> cache = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
apancorb: I agree, this can be ambiguous especially for future readers. Updated to use the `Instant` class for all timestamps.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
apancorb: I feel like this is simple to understand. Also the null checks have to be done before evaluating the expression either way, so extracting it into a variable wouldn't help that much. Wdyt?

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: Sounds good, updated to expose a package-private method that returns the reference to the `cache`. I also have updated the methods so that it uses this approach.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: Makes sense, but the reason I chose those names is to be in line with the vars and method names in main code. If you this is more readable, we can change the "invalid" references to "expired" across all files. wdyt?

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: Same comment as above.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+  void put_shouldNotPutInCache_whenPegoutCreationRskTxHashIsNull() {
+    Keccak256 pegoutCreationRskTxHash = null;
+
+    pegoutSignedCache.put(pegoutCreationRskTxHash);
+
+    verify(cache, never()).put(any(Keccak256.class), anyLong());
+  }
+
+  @Test
+  void put_shouldPutInCache_whenPegoutCreationRskTxHashIsNotNull() {
```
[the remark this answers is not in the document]
apancorb: Tysm!! This might have just prevented a nasty bug 😄 

If we put the same rsk tx hash, we are currently updating the timestamp value. Rendering the cache use ineffective since we will keep updating the timestamp and it will never reach an "expired" state. We should only put the rsk tx hash in the cache if it is absent:

`cache.putIfAbsent(rskTxHash, currentTimestamp)`

 Added new unit test.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Override
+  public void put(Keccak256 pegoutCreationRskTxHash) {
+    Instant currentTimestamp = Instant.now();
+
+    Optional.ofNullable(pegoutCreationRskTxHash)
+        .ifPresent(rskTxHash -> cache.putIfAbsent(rskTxHash, currentTimestamp));
+  }
+
+  private boolean isValidTimestamp(Instant currentTimestamp, Instant timestamp) {
```
marcos-iov, line 38: Do you need to receive `currentTimestamp` here as a parameter? Looks like you could just get inside the method calling `Instant.now()` and avoid checking is it's null

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    return currentTimestamp != null && timestamp != null &&
+        currentTimestamp.toEpochMilli() - timestamp.toEpochMilli() <= ttl.toMillis();
```
marcos-iov, line 40:
```suggestion
    timeInCache =  currentTimestamp.toEpochMilli() - timestamp.toEpochMilli();
    return currentTimestamp != null && timestamp != null &&
        timeInCache <= ttl.toMillis();
```
Feel free to use other names but I think that helps understand what's going on. Putting things in a variable with a name

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactoryTest.java
```diff
@@ -0,0 +1,40 @@
...
+        "Invalid pegouts signed cache TTL value in minutes supplied: %s", ttl);
+
+    IllegalArgumentException exception = assertThrows(IllegalArgumentException.class,
+        () -> PegoutSignedCacheFactory.getInstance(invalidTtl));
+    assertEquals(expectedErrorMessage, exception.getMessage());
+  }
+
+  @Test
+  void getInstance_shouldReturnSameInstance_whenCalledMultipleTimesWithValidTtl() {
+    Duration ttl = Duration.ofMinutes(10);
```
marcos-iov, line 31: So what happens if you call `getInstance` with different ttl values? Is it still the same instance? Which ttl value is used?

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+    boolean result = pegoutSignedCache.hasAlreadyBeenSigned(PEGOUT_CREATION_RSK_HASH);
+
+    assertFalse(result);
+  }
+
+  @Test
+  void hasAlreadyBeenSigned_shouldReturnFalse_whenCacheContainsInvalidTimestamp() {
+    Instant currentTimestamp = Instant.now();
+    Instant timestamp = currentTimestamp.minusMillis(Duration.ofMinutes(60).toMillis());
+    pegoutSignedCache.getCache().put(PEGOUT_CREATION_RSK_HASH, timestamp);
```
marcos-iov, line 46: I feel like this can be achieved by calling directly `pegoutSignedCache.put`. So we stick to testing using only the exposed methods of the class, like a normal use case

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+class PegoutSignedCacheImplTest {
+
+  private static final Duration DEFAULT_TTL = Duration.ofMinutes(30);
+  private static final Keccak256 PEGOUT_CREATION_RSK_HASH = TestUtils.createHash(1);
+
+  private final PegoutSignedCacheImpl pegoutSignedCache = new PegoutSignedCacheImpl(DEFAULT_TTL);
+
+  @BeforeEach
+  void setUp() {
+    pegoutSignedCache.getCache().clear();
```
marcos-iov, line 23: This could be achieved by simply recreating PegoutSignedCacheImpl object

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.time.Instant;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class PegoutSignedCacheImplTest {
+
+  private static final Duration DEFAULT_TTL = Duration.ofMinutes(30);
+  private static final Keccak256 PEGOUT_CREATION_RSK_HASH = TestUtils.createHash(1);
+
+  private final PegoutSignedCacheImpl pegoutSignedCache = new PegoutSignedCacheImpl(DEFAULT_TTL);
```
marcos-iov, line 19:
```suggestion
  private final PegoutSignedCache pegoutSignedCache = new PegoutSignedCacheImpl(DEFAULT_TTL);
```

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+    boolean result = pegoutSignedCache.hasAlreadyBeenSigned(PEGOUT_CREATION_RSK_HASH);
+
+    assertTrue(result);
+  }
+
+  @Test
+  void put_shouldNotPutInCache_whenPegoutCreationRskTxHashIsNull() {
+    Keccak256 pegoutCreationRskTxHash = null;
+
+    pegoutSignedCache.put(pegoutCreationRskTxHash);
```
marcos-iov, line 68: Should this throw an exception? It could be deceiving if somebody tries to insert a `null` value, no error is thrown but also nothing is saved. No way of knowing that something isn't right

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @VisibleForTesting
+  Map<Keccak256, Instant> getCache() {
```
marcos-iov, line 44: Left some comments in the test class. This probably isn't needed. We could stick to testing the class with the exposed methods. Or is it important to access the internal state of the cache? Maybe it's needed to check the timestamps used, let me know your thoughts

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,92 @@
...
+  void put_shouldPutInCacheOnce_whenPegoutCreationRskTxHashIsTheSame() {
+    // first insert
+    pegoutSignedCache.put(PEGOUT_CREATION_RSK_HASH);
+    Instant firstTimestamp = pegoutSignedCache.getCache().get(PEGOUT_CREATION_RSK_HASH);
+    // second insert
+    pegoutSignedCache.put(PEGOUT_CREATION_RSK_HASH);
+    Instant secondTimestamp = pegoutSignedCache.getCache().get(PEGOUT_CREATION_RSK_HASH);
+
+    // both objects should have the same reference
+    assertSame(firstTimestamp, secondTimestamp);
```
marcos-iov, line 90: Reading this test makes me wonder if the timestamps are the same because both `put` operations were called instantly one after the other. I'd like to see some sort of assertion that the current timestamp is different for the first element than for the second one. Makes sense?

Not sure what would be the best strategy though. Adding a `sleep` in between inserts is probably not very good. Any ideas?

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
```diff
@@ -0,0 +1,47 @@
...
+        ttl.toMinutes());
+    return instance;
+  }
+
+  private static void assertValidTtl(final Duration ttl) throws IllegalArgumentException {
+    if (ttl == null || ttl.isNegative() || ttl.isZero()) {
+      String message = String.format(
+          "Invalid pegouts signed cache TTL value in minutes supplied: %d",
+          ttl != null ? ttl.toMinutes() : null);
+      logger.error("[getInstance] {}", message);
```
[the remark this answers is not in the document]
apancorb: Comitted suggestion.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
julia-zack: didnt know, ok!

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
julia-zack: i would do it anyway, inside new `assertValidTtl` method

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
apancorb: Updated the method, let me know what you think

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
apancorb: I believe that yes. `isValidTimestamp` will later be called by the clenpup method from the periodic scheduler. Since it will iterate for each record calling this method I believe it would be a better idea to use the same `currentTimestamp` for each record rather than a new one each time the method is called. That was my reasoning behind it, wdyt?

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactoryTest.java
apancorb: Yes, once it is instantiated it returns the same instance as the first one. The ttl will be passed from a config file which is not known at creation time. If we want it to be parameterless then we can include a setter or init method that updates the ttl value for `PegoutSignedCacheImpl`.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: In this test it tests that both objects have the same reference so they are the same object. I have also included a new test for putting two different rsk tx hashes. In this case the two timestamps are not the same.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
julia-zack: i think thats a very good reason to get the `Instant.now()` each time.
If not, we could end up not cleaning a record that should be cleaned, right? because we will be behind real time

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @Test
+  void put_shouldPutInCacheBoth_whenPegoutCreationRskTxHashAreNotSame() {
+    // first insert
+    pegoutSignedCache.put(PEGOUT_CREATION_RSK_HASH);
+    Instant firstTimestamp = pegoutSignedCache.getCache().get(PEGOUT_CREATION_RSK_HASH);
+    // second insert
+    Keccak256 otherPegoutCreationRskTxHash = TestUtils.createHash(2);
+    pegoutSignedCache.put(otherPegoutCreationRskTxHash);
+    Instant secondTimestamp = pegoutSignedCache.getCache().get(otherPegoutCreationRskTxHash);
```
julia-zack, line 89:
```
    Instant pegoutCreationRskTxHashTimestamp = pegoutSignedCache.getCache().get(PEGOUT_CREATION_RSK_HASH);
    // second insert
    Keccak256 otherPegoutCreationRskTxHash = TestUtils.createHash(2);
    pegoutSignedCache.put(otherPegoutCreationRskTxHash);
    Instant otherPegoutCreationRskTxHashTimestamp = pegoutSignedCache.getCache().get(otherPegoutCreationRskTxHash);
```

i think is clearer that way, to know which timestamp corresponds to which pegout :)

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
nathanieliov: Why not just add one more method to know the size of the cache? With this method, I think we should be able to test the class properly. Also, I consider this method useful to avoid checking if a list of peg-outs is in the cache when the cache is empty.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
nathanieliov: Agree with @marcos-iov

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
nathanieliov: I agree with @julia-zack. I would rename it to expired timestamp. An invalid timestamp is not possible.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
nathanieliov: Agree with @marcos-iov

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
apancorb: Consider the case when a record has expired but the cleanup scheduler has not kicked in yet. If we call the method to check if it is present in the cache we would expect to return false.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
apancorb: Updated

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: Sounds good then, updated all "invalid" to "expired" wording.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: From what we discussed offline, updated to use reflection in tests. Let me know what you think.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactoryTest.java
apancorb: From what we discussed offline, we will not be using the singleton patter for the cache impl.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: As we discussed, this is not possible to achieve. The idea will be use reflection for this case.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: With the new changes not used anymore.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: Alright, updated.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
```diff
@@ -0,0 +1,47 @@
...
+  private PegoutSignedCacheFactory() {
+  }
+
+  /**
+   * Retrieves an instance of PegoutSignedCache with the specified Time To Live
+   * (TTL). If an instance does not already exist, it is lazily initialized.
+   * 
+   * @param ttl The duration for which the cache entries are considered valid.
+   * @return An instance of PegoutSignedCache.
+   * @throws IllegalArgumentException If the supplied TTL value is invalid.
```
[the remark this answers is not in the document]
apancorb: This is interesting. What are your thoughts @marcos-iov ?

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
```diff
@@ -0,0 +1,47 @@
...
+      instance = new PegoutSignedCacheImpl(ttl);
+    }
+
+    logger.info(
+        "[getInstance] Retrieved pegouts signed cache with TTL value in minutes: {}",
+        ttl.toMinutes());
+    return instance;
+  }
+
+  private static void assertValidTtl(final Duration ttl) throws IllegalArgumentException {
```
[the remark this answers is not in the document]
apancorb: Good catch, updated :)

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
```diff
@@ -0,0 +1,47 @@
+package co.rsk.federate.btcreleaseclient.cache;
+
+import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PegoutSignedCacheFactory {
```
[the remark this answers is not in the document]
apancorb: Not using a singleton patter anymore, we are creating a new instance for `CachePegoutSignedImpl` each time.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+class PegoutSignedCacheImpl implements PegoutSignedCache {
+
+  private final Map<Keccak256, Instant> cache = new ConcurrentHashMap<>();
+  private final Duration ttl;
+
+  PegoutSignedCacheImpl(Duration ttl) {
+    this.ttl = ttl;
```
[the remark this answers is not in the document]
apancorb: Delegated to Factory class, makes more sense imo

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,52 @@
...
+        .map(timestamp -> isValidTimestamp(currentTimestamp, timestamp))
+        .orElse(false);
+  }
+
+  @Override
+  public void put(Keccak256 pegoutCreationRskTxHash) {
+    Instant currentTimestamp = Instant.now();
+
+    Optional.ofNullable(pegoutCreationRskTxHash)
+        .ifPresent(rskTxHash -> cache.putIfAbsent(rskTxHash, currentTimestamp));
```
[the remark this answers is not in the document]
apancorb: Sounds good to me, updated the contract.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
apancorb: I have updated for the method to receive only the timestamp cached. Be advised that when called by the scheduler this will compare each record with a new timestamp making it non consistent per round of execution.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: What exception should we throw in this case?

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  private boolean isValidTimestamp(Instant currentTimestamp, Instant timestamp) {
+    if (currentTimestamp == null || timestamp == null) {
```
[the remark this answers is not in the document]
apancorb: Updated the method with new logic. Let me know what you think.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCache.java
```diff
@@ -0,0 +1,26 @@
...
+
+public interface PegoutSignedCache {
+
+  /**
+   * Checks if the specified RSK transaction hash for pegout creation has already
+   * been signed.
+   * 
+   * @param pegoutCreationRskTxHash The Keccak256 hash of the RSK transaction for
+   *                                pegout creation.
+   * @return {@code true} if the transaction hash has already been signed,
```
julia-zack, line 13: *if that transaction has already been signed
a bit nit, but more accurate to say that what was signed was a transaction (not a transaction hash)

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,39 @@
...
+        .orElse(false);
+  }
+
+  @Override
+  public void putIfAbsent(Keccak256 pegoutCreationRskTxHash) {
+    Optional.ofNullable(pegoutCreationRskTxHash)
+        .ifPresent(rskTxHash -> cache.putIfAbsent(rskTxHash, Instant.now()));
+  }
+
+  private boolean hasTimestampExpired(Instant timestampInCache) {
```
julia-zack, line 33: i'd put this method right after `hasAlreadyBeenSigned` since its being used there. I find it more readable that way

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+  void hasAlreadyBeenSigned_shouldReturnFalse_whenPegoutCreationRskTxHashIsNotCached() {
+    boolean result = pegoutSignedCache.hasAlreadyBeenSigned(PEGOUT_CREATION_RSK_HASH);
+
+    assertFalse(result);
+  }
+
+  @Test
+  void hasAlreadyBeenSigned_shouldReturnFalse_whenCacheContainsExpiredTimestamp() {
+    Instant currentTimestamp = Instant.now();
+    Instant timestamp = currentTimestamp.minusMillis(Duration.ofMinutes(60).toMillis());
```
julia-zack, line 54:
```
Instant moreTimeThanTtl = Duration.ofMinutes(60).toMillis();
Instant expiredTimestamp = currentTimestamp.minusMillis(moreTimeThanTtl);
```

(feel free to change the name, i couldn't think of a better one)

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    boolean result = pegoutSignedCache.hasAlreadyBeenSigned(PEGOUT_CREATION_RSK_HASH);
+
+    assertFalse(result);
+  }
+
+  @Test
+  void hasAlreadyBeenSigned_shouldReturnTrue_whenCacheContainsNotExpiredTimestamp() {
+    Instant currentTimestamp = Instant.now();
+    Instant timestamp = currentTimestamp.minusMillis(Duration.ofMinutes(10).toMillis());
```
julia-zack, line 65: same here

```
Instant lessTimeThanTtl = Duration.ofMinutes(10).toMillis();
Instant timestamp = currentTimestamp.minusMillis(moreTimeThanTtl);
```

(again feel free to change the name)

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+    assertTrue(result);
+  }
+
+  @Test
+  void putIfAbsent_shouldNotPutInCache_whenPegoutCreationRskTxHashIsNull() {
+    Keccak256 pegoutCreationRskTxHash = null;
+
+    pegoutSignedCache.putIfAbsent(pegoutCreationRskTxHash);
+
+    assertFalse(cache.containsKey(PEGOUT_CREATION_RSK_HASH));
```
julia-zack, line 79: shouldn't you check for `pegoutCreationRskTxHash`?

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+  void putIfAbsent_shouldPutInCacheBoth_whenPegoutCreationRskTxHashAreNotSame() {
+    // first insert
+    pegoutSignedCache.putIfAbsent(PEGOUT_CREATION_RSK_HASH);
+    Instant pegoutCreationRskTxHashTimestamp = cache.get(PEGOUT_CREATION_RSK_HASH);
+    // second insert
+    Keccak256 otherPegoutCreationRskTxHash = TestUtils.createHash(2);
+    pegoutSignedCache.putIfAbsent(otherPegoutCreationRskTxHash);
+    Instant otherPegoutCreationRskTxHashTimestamp = cache.get(otherPegoutCreationRskTxHash);
+
+    assertNotSame(pegoutCreationRskTxHashTimestamp, otherPegoutCreationRskTxHashTimestamp);
```
julia-zack, line 99: to stick with the test name i would first assert that both rsk hashes are in the cache.
I know that that's implied when getting the timestamp using their keys, but I think when you read the test name you are expecting this to happen... checking the timestamps are different is an addition to the test, not its "main purpose".
Wdyt?

```
    // first insert
    pegoutSignedCache.putIfAbsent(PEGOUT_CREATION_RSK_HASH);
    
    // second insert
    Keccak256 otherPegoutCreationRskTxHash = TestUtils.createHash(2);
    pegoutSignedCache.putIfAbsent(otherPegoutCreationRskTxHash);

   assertTrue(cache.containsKey(PEGOUT_CREATION_RSK_HASH));
   assertTrue(cache.containsKey(otherPegoutCreationRskTxHash));   


    // also check that they have different timestamps
    Instant pegoutCreationRskTxHashTimestamp = cache.get(PEGOUT_CREATION_RSK_HASH);
    Instant otherPegoutCreationRskTxHashTimestamp = cache.get(otherPegoutCreationRskTxHash);

    assertNotSame(pegoutCreationRskTxHashTimestamp, otherPegoutCreationRskTxHashTimestamp);
```

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  @Test
+  void putIfAbsent_shouldPutInCacheOnce_whenPegoutCreationRskTxHashIsTheSame() {
+    // first insert
+    pegoutSignedCache.putIfAbsent(PEGOUT_CREATION_RSK_HASH);
+    Instant pegoutCreationRskTxHashTimestamp = cache.get(PEGOUT_CREATION_RSK_HASH);
+    // second insert
+    pegoutSignedCache.putIfAbsent(PEGOUT_CREATION_RSK_HASH);
+    Instant otherPegoutCreationRskTxHashTimestamp = cache.get(PEGOUT_CREATION_RSK_HASH);
```
julia-zack, line 109: `Instant pegoutCreationRskTxHashTimestamp2 = cache.get(PEGOUT_CREATION_RSK_HASH);`

the "other" at the beginning makes me think that you are talking about another pegout and not about another timestamp (overall because above is being used that way)

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: Good catch! :)

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
julia-zack: a good idea could be checking for cache size, as you did [here](https://github.com/rsksmart/powpeg-node/blob/25acf0f329e2279715b9d44da1d255d3f9cbc1cb/src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java#L86)

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    return Optional.ofNullable(timestampInCache)
+      .map(timestamp -> Instant.now().toEpochMilli() - timestamp.toEpochMilli())
+      .map(timeCachedInMillis -> timeCachedInMillis <= ttl.toMillis())
```
nathanieliov, line 36:
```suggestion
      .map(timeElapsed -> timeElapsed.compareTo(ttl))
      .map(comparisonResult -> comparisonResult < 0)
```

Wdyt?

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+  PegoutSignedCacheImpl(Duration ttl) {
+    this.ttl = ttl;
+  }
+
+  @Override
+  public boolean hasAlreadyBeenSigned(Keccak256 pegoutCreationRskTxHash) {
+    return Optional.ofNullable(pegoutCreationRskTxHash)
+        .map(cache::get)
+        .map(this::hasTimestampExpired)
```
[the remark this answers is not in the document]
apancorb: I changed the name to `hasTimestampNotExpired`

marcos-iov (approved):

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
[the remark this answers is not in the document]
marcos-iov: https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html

> Here's the bottom line guideline: If a client can reasonably be expected to recover from an exception, make it a checked exception. If a client cannot do anything to recover from the exception, make it an unchecked exception.

Not sure we can recover from an invalid cache ttl value. We can't start the cache and we can't process peg-outs. IllegalArgumentExcepiton sounds good to me.

https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html
> Thrown to indicate that a method has been passed an illegal or inappropriate argument.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
marcos-iov: IllegalArgumentException? I think it's better thp throw an exception than to simply ignore the fact that somebody tried to insert a null hash. That could mean something is wrong

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheFactory.java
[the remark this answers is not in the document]
apancorb: I agree with @marcos-iov . We shouldn't allow the process to begin until a valid ttl is found.

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: I agree. I would rather throw an NPE.

From the javadoc of the cache `putIfAbsent` method:

> if the specified key or value is null, and this map does not permit null keys or values

I think we can use this exception since that is what the underlying cache is doing. Anyways in the meantime I have added the exception and updated unit test. Wdyt?

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: Not sure, checking if the timestamps are not the same (i.e. don't hold the same reference) is the main purpose of the test. The other asserts you mention are already covered in their own test even though as you mention it is implied and could be placed as well. Ideally we should test one thing only per test, "ideally".

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: Something similar is what I originally had, will update to this.

src/main/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCache.java
apancorb: I changed the wording, let me know what you think

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
```diff
@@ -0,0 +1,130 @@
...
+    Instant notExpiredTimestamp = currentTimestamp.minusMillis(Duration.ofMinutes(10).toMillis());
+    cache.put(PEGOUT_CREATION_RSK_HASH, notExpiredTimestamp);
+
+    boolean result = pegoutSignedCache.hasAlreadyBeenSigned(PEGOUT_CREATION_RSK_HASH);
+
+    assertTrue(result);
+  }
+
+  @Test
+  void putIfAbsent_shouldThrowNullPointerException_whenPegoutCreationRskTxHashIsNull() {
```
marcos-iov, line 90:
```suggestion
  void putIfAbsent_shouldThrowIllegalArgumentException_whenPegoutCreationRskTxHashIsNull() {
```
:)

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
julia-zack: i agree with what you are saying, but when i read the test name i expect to check if both values are in the cache

but leave it that way if you think is clear enough! no worries

julia-zack (approved):
approved anyways :)))

nathanieliov (approved):
Approved. Regarding the `hasTimestampNotExpired` implementation, I will leave my suggestion discussion open in case you decide to implement it.

Well done Antonio!

src/test/java/co/rsk/federate/btcreleaseclient/cache/PegoutSignedCacheImplTest.java
apancorb: Thanks, updated!

PR #269, opened 2024-06-06
Refactor PowHSMBookkeepingConfig and Add missing test coverage 

apancorb:
### Summary

* Refactor of `PowHSMBokkepingConfig` class to be more consistent with the rest of the config classes in the `powpeg-node` project.
* Add missing tests to `PowHSMBookeepingConfigTest` for other exposed config related methods.

In the next PR we will change the difficultyTarget method to take the hsmVersion and return the needed difficultyTraget. 

*No new main logic is introduced*

### Test Plan

Unit tests

### DIFF
--- a/src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
+++ b/src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
@@ -9,7 +9,7 @@ public class PowHSMBookkeepingConfig {
 
     private static final Logger logger = LoggerFactory.getLogger(PowHSMBookkeepingConfig.class);
 
-    enum NetworkDifficultyCap {
+    public enum NetworkDifficultyCap {
         MAINNET(new BigInteger("7000000000000000000000")),
         TESTNET(BigInteger.valueOf(1000000000000000L)),
         REGTEST(BigInteger.valueOf(20L));
@@ -20,7 +20,7 @@ public class PowHSMBookkeepingConfig {
             this.difficultyCap = difficultyCap;
         }
 
-        BigInteger getDifficultyCap() {
+        public BigInteger getDifficultyCap() {
             return difficultyCap;
         }
     }
--- a/src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
+++ b/src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
@@ -1,13 +1,12 @@
 package co.rsk.federate.config;
 
 import co.rsk.bitcoinj.core.NetworkParameters;
+import com.typesafe.config.Config;
 import java.math.BigInteger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class PowHSMBookkeepingConfig {
-
-    private static final Logger logger = LoggerFactory.getLogger(PowHSMBookkeepingConfig.class);
 
     public enum NetworkDifficultyCap {
         MAINNET(new BigInteger("7000000000000000000000")),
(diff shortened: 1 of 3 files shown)


src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
```diff
@@ -1,87 +1,86 @@
...
-            this.maxChunkSizeToHsm = signerConfig.getConfig().getInt("bookkeeping.maxChunkSizeToHsm");
+    public enum NetworkDifficultyCap {
+        MAINNET(new BigInteger("7000000000000000000000")),
+        TESTNET(BigInteger.valueOf(1000000000000000L)),
+        REGTEST(BigInteger.valueOf(20L));
+
+        private final BigInteger difficultyCap;
+
+        NetworkDifficultyCap(BigInteger difficultyCap) {
+            this.difficultyCap = difficultyCap;
```
julia-zack, line 20: would it make sense to have this enum in a separated class?

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
```diff
@@ -1,87 +1,86 @@
...
+        public BigInteger getDifficultyCap() {
+            return difficultyCap;
         }
     }
 
+    private static final BigInteger DEFAULT_DIFFICULTY_TARGET = BigInteger.valueOf(3);
+    private static final boolean DEFAULT_STOP_BOOKKEPING_SCHEDULER = false;
+    private static final int DEFAULT_MAX_AMOUNT_BLOCK_HEADERS = 7;
+    private static final int DEFAULT_MAX_CHUNK_SIZE = 10;
+    private static final long DEFAULT_INFORMER_INTERVAL = 2_000;
```
julia-zack, line 32: same here?

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
```diff
@@ -1,87 +1,86 @@
...
+    public PowHSMBookkeepingConfig(SignerConfig signerConfig, String networkParameter) {
+        this.signerConfig = signerConfig;
+        this.networkParameter = networkParameter;
+    }
+
     public BigInteger getDifficultyTarget() {
-        return difficultyTarget;
+        return signerConfig.getConfig().hasPath("bookkeeping.difficultyTarget")
+            ? new BigInteger(signerConfig.getConfig().getString("bookkeeping.difficultyTarget"))
+            : DEFAULT_DIFFICULTY_TARGET;
```
julia-zack, line 45: what do you think about refactoring these methods?
To have something like

```
public BigInteger getDifficultyTarget() {

     if (!hasDifficultyTargetConfig(signerConfig)) {
          return DEFAULT_DIFFICULTY_TARGET;
     }

     return new BigInteger(signerConfig.getConfig().getString("bookkeeping.difficultyTarget"));
}

private boolean hasDifficultyTargetConfig(signerConfig) {
     return signerConfig.getConfig().hasPath("bookkeeping.difficultyTarget");
}
```

src/test/java/co/rsk/federate/config/PowHSMBookkeepingConfigTest.java
```diff
@@ -1,57 +1,145 @@
...
+    void getInformerInterval_whenCustomConfigNotAvailable_shouldReturnDefaultConfig() {
+        when(config.hasPath("bookkeeping.informerInterval")).thenReturn(false);
+
+        assertEquals(DEFAULT_INFORMER_INTERVAL, powHsmBookkeepingConfig.getInformerInterval());
+    }
+
+    @Test
+    void isStoppingBookkepingScheduler_whenCustomConfigAvailable_shouldReturnCustomConfig() {
+        boolean customIsStoppingInformerInterval = !DEFAULT_STOP_BOOKKEPING_SCHEDULER; 
+        when(config.hasPath("bookkeeping.stopBookkeepingScheduler")).thenReturn(true);
```
julia-zack, line 106: these is a bit confusing for me, wdyt about doing (something like) this:

```
boolean customIsStoppingInformerInterval = !DEFAULT_STOP_BOOKKEPING_SCHEDULER; // default is set as false 
when(config.hasPath("bookkeeping.stopBookkeepingScheduler")).thenReturn(customIsStoppingInformerInterval);
```

src/test/java/co/rsk/federate/config/PowHSMBookkeepingConfigTest.java
```diff
@@ -1,57 +1,145 @@
...
+        PowHSMBookkeepingConfig powHsmBookkeepingConfig = new PowHSMBookkeepingConfig(signerConfig, networkId);
+
+        assertEquals(expectedNetworkDifficultyCap.getDifficultyCap(),
+            powHsmBookkeepingConfig.getDifficultyCap());
+    }
+
+    @Test
+    void getDifficultyCap_whenGivenInvalidNetwork_showThrowIllegalArgumentException() {
+        PowHSMBookkeepingConfig powHsmBookkeepingConfig =
+            new PowHSMBookkeepingConfig(signerConfig, NetworkParameters.ID_UNITTESTNET);
```
julia-zack, line 132: i'd call it `UNKNOWN_NETWORK` or similar, it was a bit confusing for me at first

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
apancorb: Since it is only being consumed inside the scope of this class it can be left here for now. If it starts being consumed by two or more classes then we can extract it.

src/test/java/co/rsk/federate/config/PowHSMBookkeepingConfigTest.java
apancorb: I agree, but that class is out of the scope of this change :)

src/test/java/co/rsk/federate/config/PowHSMBookkeepingConfigTest.java
```diff
@@ -1,57 +1,145 @@
...
+
+        assertThrows(IllegalArgumentException.class,
+            () -> powHsmBookkeepingConfig.getDifficultyCap());
```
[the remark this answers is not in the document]
apancorb: Updated, thanks

src/test/java/co/rsk/federate/config/PowHSMBookkeepingConfigTest.java
julia-zack: haha you right, my bad

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
julia-zack: is a good point, wdyt about this

```
public BigInteger getDifficultyTarget() {

     boolean hasDifficultyTargetConfig = signerConfig.getConfig().hasPath("bookkeeping.difficultyTarget");

     if (!hasDifficultyTargetConfig) {
          return DEFAULT_DIFFICULTY_TARGET;
     }

     return new BigInteger(signerConfig.getConfig().getString("bookkeeping.difficultyTarget"));
}
```

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
```diff
@@ -1,87 +1,86 @@
...
-            default:
-                String message = "Invalid network specified for the Bookkeeping Config: " + networkParameter;
-                logger.error(message);
-                throw new IllegalArgumentException(message);
+        public BigInteger getDifficultyCap() {
+            return difficultyCap;
         }
     }
 
+    private static final BigInteger DEFAULT_DIFFICULTY_TARGET = BigInteger.valueOf(3);
```
marcos-iov, line 28: This should be gone, right? Is that part of upcoming changes?

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
```diff
@@ -1,87 +1,86 @@
...
     public int getMaxChunkSizeToHsm() {
-        return maxChunkSizeToHsm;
+        return signerConfig.getConfig().hasPath("bookkeeping.maxChunkSizeToHsm")
+            ? signerConfig.getConfig().getInt("bookkeeping.maxChunkSizeToHsm")
+            : DEFAULT_MAX_CHUNK_SIZE;
     }
 
     public BigInteger getDifficultyCap() {
-        return difficultyCap;
+        switch (networkParameter) {
```
marcos-iov, line 73: I believe the network can also be obtained from the device for HSM >= 4

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
```diff
@@ -1,87 +1,86 @@
...
+    private final String networkParameter;
+
+    public PowHSMBookkeepingConfig(SignerConfig signerConfig, String networkParameter) {
+        this.signerConfig = signerConfig;
+        this.networkParameter = networkParameter;
+    }
+
     public BigInteger getDifficultyTarget() {
-        return difficultyTarget;
+        return signerConfig.getConfig().hasPath("bookkeeping.difficultyTarget")
```
marcos-iov, line 43: Should we put the paths in an enum maybe? Looks like we are repeating each of them at least twice, that's a bit error prone.

Perhaps the enum could have 2 attributes, path and default value

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
```diff
@@ -1,87 +1,86 @@
...
+    private static final boolean DEFAULT_STOP_BOOKKEPING_SCHEDULER = false;
+    private static final int DEFAULT_MAX_AMOUNT_BLOCK_HEADERS = 7;
+    private static final int DEFAULT_MAX_CHUNK_SIZE = 10;
+    private static final long DEFAULT_INFORMER_INTERVAL = 2_000;
+
+    private final SignerConfig signerConfig;
+    private final String networkParameter;
+
+    public PowHSMBookkeepingConfig(SignerConfig signerConfig, String networkParameter) {
+        this.signerConfig = signerConfig;
```
marcos-iov, line 38: Looks like we always access `singerConfi,getConfig`. So maybe we shpuld simply store the Config object?

```suggestion
        this.signerConfig = signerConfig.getConfig();
```

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
apancorb: IMO these snippets are just ternary operators. Unless the rest agrees, I would leave as is.

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
apancorb: Yep, this will be done in the next PR. This is just a refactor before introducing "new" changes, so it easier to review.

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
apancorb: Nice, I will have this in mind for the next PR.

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
apancorb: The thing is the default value can be a string, int, long, boolean etc. An enum cannot have type params, so unless we do casting there is no way to store the default values.

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
apancorb: Yes I originally was going to do it that way. But since SignerConfig was more "powHsm" related I left as is. But I agree, I have updated it.

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
```diff
@@ -1,87 +1,96 @@
...
-                logger.error(message);
-                throw new IllegalArgumentException(message);
+        public BigInteger getDifficultyCap() {
+            return difficultyCap;
         }
     }
 
+    private static final Logger logger = LoggerFactory.getLogger(PowHSMBookkeepingConfig.class);
+
+    private static final String DIFFICULTY_TARGET_PATH = "bookkeeping.difficultyTarget";
```
marcos-iov, line 29: Maybe for another story, but I'd consider moving the paths to an enum on a separate file. To clean up this class a bit.

Same for NetworkDifficultyCap enum, in a separate file it will simplify this class.

Default values make sense to keep them here I think

apancorb: I think we can make this in another story, so we can also put the paths present in `src/main/java/co/rsk/federate/config/FedNodeSystemProperties.java` in this enum as well (this way we consolidate a bit). We can come up with a better overall design and try to see how we can keep the defaults there. We can also include NetworkDifficultyCap there. I can create this story if you want.

marcos-iov (approved):

PR #270, opened 2024-06-06
Adapt powpeg-node to be compatible with latest Federation refactors

marcos-iov:
- Apply changes in powpeg node classes and tests to be compatible with the latest changes implemented in `federation-support-refactor-integration` branch in rskj

### DIFF
--- a/src/test/java/co/rsk/federate/adapter/ThinConverterTest.java
+++ b/src/test/java/co/rsk/federate/adapter/ThinConverterTest.java
@@ -13,18 +13,16 @@ class ThinConverterTest {
 
     private static int nhash = 0;
 
-    public static org.bitcoinj.core.Sha256Hash createOriginalHash() {
+    private org.bitcoinj.core.Sha256Hash createOriginalHash() {
         byte[] bytes = new byte[32];
         bytes[0] = (byte) nhash++;
-        org.bitcoinj.core.Sha256Hash hash = org.bitcoinj.core.Sha256Hash.wrap(bytes);
-        return hash;
+        return org.bitcoinj.core.Sha256Hash.wrap(bytes);
     }
 
-    public static co.rsk.bitcoinj.core.Sha256Hash createThinHash() {
+    private co.rsk.bitcoinj.core.Sha256Hash createThinHash() {
         byte[] bytes = new byte[32];
         bytes[0] = (byte) nhash++;
-        co.rsk.bitcoinj.core.Sha256Hash hash = co.rsk.bitcoinj.core.Sha256Hash.wrap(bytes);
-        return hash;
+        return co.rsk.bitcoinj.core.Sha256Hash.wrap(bytes);
     }
 
     @Test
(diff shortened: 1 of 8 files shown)


src/test/java/co/rsk/federate/adapter/ThinConverterTest.java
```diff
@@ -13,18 +13,16 @@ class ThinConverterTest {
 
     private static int nhash = 0;
 
-    public static org.bitcoinj.core.Sha256Hash createOriginalHash() {
+    private org.bitcoinj.core.Sha256Hash createOriginalHash() {
```
julia-zack, line 16: just a thought, could we just import this class? I think it adds unnecessary noise like this

IMHO, it will look much clearer to see `private Sha256Hash createOriginalHash()`

src/test/java/co/rsk/federate/adapter/ThinConverterTest.java
```diff
@@ -34,12 +32,12 @@ void toThinInstanceStoredBlock() {
         int height = 200;
         org.bitcoinj.core.Block originalBlock = new org.bitcoinj.core.Block(params, 1, createOriginalHash(), createOriginalHash(), 1000, 2000, 3000, new ArrayList<>());
         org.bitcoinj.core.StoredBlock originalStoredBlock = new org.bitcoinj.core.StoredBlock(originalBlock, chainWork, height);
-        co.rsk.bitcoinj.core.StoredBlock thinStoredBlock = ThinConverter.toThinInstance(originalStoredBlock, BridgeRegTestConstants.getInstance());
+        co.rsk.bitcoinj.core.StoredBlock thinStoredBlock = ThinConverter.toThinInstance(originalStoredBlock, new BridgeRegTestConstants());
```
julia-zack, line 35: same here

src/test/java/co/rsk/federate/adapter/ThinConverterTest.java
marcos-iov: One is creating a hash from bitcoinj-core and the other one from bitcoinj-thin. That's why the have the full package name, to distinguish them

jeremy-then: I agree. And it is already imported.

src/test/java/co/rsk/federate/adapter/ThinConverterTest.java
jeremy-then: I agree. It doesn't collide with any other class with the name name in this file, so we can simply put `StoredBlock`.

jeremy-then (approved):
LGTM.

src/test/java/co/rsk/federate/adapter/ThinConverterTest.java
marcos-iov: We have 
```
org.bitcoinj.core.StoredBlock originalStoredBlock
co.rsk.bitcoinj.core.StoredBlock thinStoredBlock
```
It's important to distinguish between the 2. ThinConverter class is used to convert from bitcoinj-thin to bitcoinj-core and viceversa

PR #271, opened 2024-06-11
Dynamically load the difficultyTarget config value from PowHSM blockchainParameters RPC

apancorb:
### Summary

As it stands right now, the powpeg node allows a configuration field, signers.btc.bookkeeping.difficultyTarget, to define the minimum amount collected by the next segmented chain to pass for the advanceBlockchain RPC. The algorithm to find the next valid chain is implemented [here](https://github.com/rsksmart/powpeg-node/blob/master/src/main/java/co/rsk/federate/signing/hsm/advanceblockchain/ConfirmedBlocksProvider.java#L34). 

In reality this value does not need to be approximated and configured since the PowHSM exposes this value by the blockchainParameters RPC. The idea is to use this value by default, and if the custom configured value exists then do an override.

We can use this method to replace the config sent in the powpeg node, only for HSM >= 3. For HSM2 we need the values to be provided in the config.

### Requirements

Remove de default config value for difficultyTarget since it might be misleading

For HSM <= 2, if the value is not present in the config value do not let the node to start up by throwing an exception. Else continue to operate as usual.

For HSM >= 3, if the value is not present in the config, use the value provided by the HSM. In other words, extract the minimum_difficulty value at start time by calling blockchainParameters if there is no custom value configured. If the call fails as well, do not let the node to start up.

### DIFF
--- a/src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
+++ b/src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
@@ -2,11 +2,10 @@ package co.rsk.federate.config;
 
 import co.rsk.bitcoinj.core.NetworkParameters;
 import co.rsk.federate.signing.hsm.HSMClientException;
+import co.rsk.federate.signing.hsm.HSMUnsupportedTypeException;
 import co.rsk.federate.signing.hsm.client.HSMBookkeepingClient;
-import co.rsk.federate.signing.hsm.message.PowHSMBlockchainParameters;
 import com.typesafe.config.Config;
 import java.math.BigInteger;
-import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


nathanieliov (approved):

src/test/java/co/rsk/federate/config/PowHSMBookkeepingConfigTest.java
```diff
@@ -46,19 +52,50 @@ void setup() {
...
+        when(hsmClient.getVersion()).thenReturn(2);
+        when(config.getString(DIFFICULTY_TARGET_PATH)).thenReturn(null);
+
+        assertThrows(IllegalStateException.class,
+            () -> powHsmBookkeepingConfig.getDifficultyTarget(hsmClient));
+    }
+
+    @Test
+    void getDifficultyTarget_whenPowHSMVersionIsGreaterThanTwo_shouldRetriveConfigValueFromPowHSM() throws Exception {
+        BigInteger difficultyTarget = BigInteger.valueOf(1L);
```
jeremy-then, line 74: nit: `expectedDifficultyTarget`? This way, when we read

`assertEquals(expectedDifficultyTarget, powHsmBookkeepingConfig.getDifficultyTarget(hsmClient));`

it would be easier.

jeremy-then (approved):
LGTM.

src/test/java/co/rsk/federate/config/PowHSMBookkeepingConfigTest.java
apancorb: Sounds good, updated!

src/main/java/co/rsk/federate/config/PowHSMBookkeepingConfig.java
```diff
@@ -49,10 +52,36 @@ public PowHSMBookkeepingConfig(SignerConfig signerConfig, String networkParamete
...
+              .orElseThrow(() -> 
+                  new IllegalStateException("Unable to read difficulty target from configuration file"));
+      }
+
+      logger.trace("[getDifficultyTarget] Retrieve minimum difficulty target from the PowHSM");
+
+      return Optional.ofNullable(hsmClient.getBlockchainParameters())
+          .map(PowHSMBlockchainParameters::getMinimumDifficulty)
+          .orElseThrow(() ->
+              new IllegalStateException("Unable to retrieve difficulty target using PowHSM"));
```
marcos-iov, line 84:
```suggestion
     try {
          logger.trace("[getDifficultyTarget] Retrieve minimum difficulty target from the PowHSM");
          
      return Optional.ofNullable(hsmClient.getBlockchainParameters())
          .map(PowHSMBlockchainParameters::getMinimumDifficulty)
     } catch(HSMUnsupportedTypeException) (
         logger(try get the value from config file)
          return Optional.ofNullable(signerConfig.getString(DIFFICULTY_TARGET_PATH))
              .map(BigInteger::new)
              .orElseThrow(() -> 
                  new IllegalStateException("Unable to read difficulty target from configuration file"));
      } 
```
What do you think?

apancorb: I agree, I have updated the method. I have also added in the catch statement the `IllegaStateException` as a fallback in case the call to the PowHSM fails with an invalid value of some sort.

marcos-iov (approved):

PR #272, opened 2024-06-12
Update ReleaseCreationInformation class to decode outpoints from receipt

nathanieliov:
## Description

- Update the ReleaseCreationInformation class to include one new properties, List<Coin> utxoOutpointValues. Get the pegoutTransactionCreatedSignatureTopic value from BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent().encodeSignature().
- Initialize the utxoOutpointValues list to an empty list like this.outpointValues = Collections.emptyList();
- Add a new List<Coin> getUtxoOutpointValues() method.
- Add a new void decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) method and save the outpoint values list in the utxoOutpointValues state variable in a Collections.unmodifiableList structure. Do a lazy loading when the getUtxoOutpointValues() method is called; if the list is null, call decodeUtxoOutpointValues.
- Add new tests for ReleaseCreationInformation class.
- Do some refactors to existing test cases of ReleaseCreationInformationGetterTest

## Motivation and Context

To add support to HSM segwit to the powpeg node

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:
- [x] My change requires a change to the documentation.

### DIFF
--- a/src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
+++ b/src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
@@ -68,7 +68,6 @@ public class ReleaseCreationInformation {
             .map(LogInfo::getData)
             .map(this::decodePegoutTransactionEventData)
             .map(UtxoUtils::decodeOutpointValues)
-            .map(Collections::unmodifiableList)
             .orElse(Collections.emptyList());
     }
 
@@ -82,8 +81,7 @@ public class ReleaseCreationInformation {
 
     private byte[] decodePegoutTransactionEventData(byte[] pegoutCreatedTransactionEventData) {
         Function pegoutTransactionCreatedEvent = BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent();
-        byte[] serializedOutpointValues = (byte[]) pegoutTransactionCreatedEvent.decodeEventData(pegoutCreatedTransactionEventData)[0];
-        return serializedOutpointValues;
+        return (byte[]) pegoutTransactionCreatedEvent.decodeEventData(pegoutCreatedTransactionEventData)[0];
     }
 
     public Block getPegoutCreationBlock() {
@@ -113,6 +111,6 @@ public class ReleaseCreationInformation {
     }
 
     public List<Coin> getUtxoOutpointValues() {
-        return utxoOutpointValues;
+        return Collections.unmodifiableList(utxoOutpointValues);
     }
 }
--- a/src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
+++ b/src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
@@ -14,6 +14,7 @@ import org.ethereum.core.TransactionReceipt;
 import org.ethereum.vm.LogInfo;
 
 public class ReleaseCreationInformation {
+
     private final Block pegoutCreationBlock;
     private final TransactionReceipt transactionReceipt;
     private final Keccak256 pegoutCreationRskTxHash;
@@ -22,11 +23,10 @@ public class ReleaseCreationInformation {
     private final List<Coin> utxoOutpointValues;
 
     /**
-     *
-     * @param pegoutCreationBlock                 The rsk block where pegout was created
-     * @param transactionReceipt    The rsk transaction receipt where pegout was created
-     * @param pegoutCreationRskTxHash      The rsk transaction hash where the pegout was created
-     * @param pegoutBtcTx        The BTC transaction to sign
+     * @param pegoutCreationBlock     The rsk block where pegout was created
+     * @param transactionReceipt      The rsk transaction receipt where pegout was created
+     * @param pegoutCreationRskTxHash The rsk transaction hash where the pegout was created
+     * @param pegoutBtcTx             The BTC transaction to sign
      **/
     public ReleaseCreationInformation(
         Block pegoutCreationBlock,
@@ -34,16 +34,17 @@ public class ReleaseCreationInformation {
         Keccak256 pegoutCreationRskTxHash,
         BtcTransaction pegoutBtcTx
     ) {
-        this(pegoutCreationBlock, transactionReceipt, pegoutCreationRskTxHash, pegoutBtcTx, pegoutCreationRskTxHash);
+        this(pegoutCreationBlock, transactionReceipt, pegoutCreationRskTxHash, pegoutBtcTx,
+            pegoutCreationRskTxHash);
     }
 
     /**
-     *
-     * @param pegoutCreationBlock                 The rsk block where the pegout was created
-     * @param transactionReceipt    The rsk transaction receipt where the pegout was created
-     * @param pegoutCreationRskTxHash      The rsk transaction hash where the pegout was created
-     * @param pegoutBtcTx        The BTC transaction to sign
-     * @param pegoutConfirmationRskTxHash  The rsk transaction hash where the pegout was confirmed to be signed
+     * @param pegoutCreationBlock         The rsk block where the pegout was created
+     * @param transactionReceipt          The rsk transaction receipt where the pegout was created
+     * @param pegoutCreationRskTxHash     The rsk transaction hash where the pegout was created
+     * @param pegoutBtcTx                 The BTC transaction to sign
+     * @param pegoutConfirmationRskTxHash The rsk transaction hash where the pegout was confirmed to
+     *                                    be signed
      **/
     public ReleaseCreationInformation(
         Block pegoutCreationBlock,
@@ -58,7 +59,7 @@ public class ReleaseCreationInformation {
         this.pegoutBtcTx = pegoutBtcTx;
         this.pegoutConfirmationRskTxHash = pegoutConfirmationRskTxHash;
 
-        this.utxoOutpointValues = this.decodeUtxoOutpointValues(transactionReceipt);
+        this.utxoOutpointValues = decodeUtxoOutpointValues(transactionReceipt);
     }
 
     private List<Coin> decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
@@ -81,7 +82,8 @@ public class ReleaseCreationInformation {
 
     private byte[] decodePegoutTransactionEventData(byte[] pegoutCreatedTransactionEventData) {
         Function pegoutTransactionCreatedEvent = BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent();
-        return (byte[]) pegoutTransactionCreatedEvent.decodeEventData(pegoutCreatedTransactionEventData)[0];
+        return (byte[]) pegoutTransactionCreatedEvent.decodeEventData(
+            pegoutCreatedTransactionEventData)[0];
     }
 
     public Block getPegoutCreationBlock() {
--- a/src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
+++ b/src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
@@ -8,7 +8,6 @@ import co.rsk.peg.bitcoin.UtxoUtils;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.Optional;
 import org.ethereum.core.Block;
 import org.ethereum.core.CallTransaction.Function;
 import org.ethereum.core.TransactionReceipt;
@@ -20,7 +19,7 @@ public class ReleaseCreationInformation {
     private final Keccak256 pegoutCreationRskTxHash;
     private final BtcTransaction pegoutBtcTx;
     private final Keccak256 pegoutConfirmationRskTxHash;
-    private List<Coin> utxoOutpointValues = Collections.emptyList();
+    private final List<Coin> utxoOutpointValues;
 
     /**
      *
@@ -59,28 +58,32 @@ public class ReleaseCreationInformation {
         this.pegoutBtcTx = pegoutBtcTx;
         this.pegoutConfirmationRskTxHash = pegoutConfirmationRskTxHash;
 
-        this.decodeUtxoOutpointValues(transactionReceipt);
+        this.utxoOutpointValues = this.decodeUtxoOutpointValues(transactionReceipt);
     }
 
-    private void decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
-        List<LogInfo> logs =  transactionReceipt.getLogInfoList();
+    private List<Coin> decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
+        return transactionReceipt.getLogInfoList().stream()
+            .filter(this::isPegoutTransactionCreatedLog)
+            .findFirst()
+            .map(LogInfo::getData)
+            .map(this::decodePegoutTransactionEventData)
+            .map(UtxoUtils::decodeOutpointValues)
+            .map(Collections::unmodifiableList)
+            .orElse(Collections.emptyList());
+    }
 
+    private boolean isPegoutTransactionCreatedLog(LogInfo log) {
         Function pegoutTransactionCreatedEvent = BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent();
         final byte[] pegoutTransactionCreatedSignatureTopic = pegoutTransactionCreatedEvent.encodeSignatureLong();
 
-        Optional<LogInfo> pegoutTransactionCreatedLog = logs.stream().filter((log ->
-            !log.getTopics().isEmpty() && Arrays.equals(log.getTopics().get(0).getData(),
-                pegoutTransactionCreatedSignatureTopic)
-        )).findFirst();
+        return !log.getTopics().isEmpty() && Arrays.equals(log.getTopics().get(0).getData(),
+            pegoutTransactionCreatedSignatureTopic);
+    }
 
-        if(!pegoutTransactionCreatedLog.isPresent()) {
-            return;
-        }
-
-        byte[] pegoutCreatedTransactionEventData = pegoutTransactionCreatedLog.get().getData();
+    private byte[] decodePegoutTransactionEventData(byte[] pegoutCreatedTransactionEventData) {
+        Function pegoutTransactionCreatedEvent = BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent();
         byte[] serializedOutpointValues = (byte[]) pegoutTransactionCreatedEvent.decodeEventData(pegoutCreatedTransactionEventData)[0];
-
-        this.utxoOutpointValues = Collections.unmodifiableList(UtxoUtils.decodeOutpointValues(serializedOutpointValues));
+        return serializedOutpointValues;
     }
 
     public Block getPegoutCreationBlock() {
--- a/src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
+++ b/src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
@@ -59,10 +59,10 @@ public class ReleaseCreationInformation {
         this.pegoutBtcTx = pegoutBtcTx;
         this.pegoutConfirmationRskTxHash = pegoutConfirmationRskTxHash;
 
-        this.utxoOutpointValues = decodeUtxoOutpointValues(transactionReceipt);
+        this.utxoOutpointValues = decodeUtxoOutpointValues();
     }
 
-    private List<Coin> decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
+    private List<Coin> decodeUtxoOutpointValues() {
         return transactionReceipt.getLogInfoList().stream()
             .filter(this::isPegoutTransactionCreatedLog)
             .findFirst()
--- a/src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
+++ b/src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
@@ -1,9 +1,18 @@
 package co.rsk.federate.signing.hsm.message;
 
 import co.rsk.bitcoinj.core.BtcTransaction;
+import co.rsk.bitcoinj.core.Coin;
 import co.rsk.crypto.Keccak256;
+import co.rsk.peg.BridgeEvents;
+import co.rsk.peg.bitcoin.UtxoUtils;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
 import org.ethereum.core.Block;
+import org.ethereum.core.CallTransaction.Function;
 import org.ethereum.core.TransactionReceipt;
+import org.ethereum.vm.LogInfo;
 
 public class ReleaseCreationInformation {
     private final Block pegoutCreationBlock;
@@ -11,6 +20,7 @@ public class ReleaseCreationInformation {
     private final Keccak256 pegoutCreationRskTxHash;
     private final BtcTransaction pegoutBtcTx;
     private final Keccak256 pegoutConfirmationRskTxHash;
+    private List<Coin> utxoOutpointValues = Collections.emptyList();
 
     /**
      *
@@ -48,6 +58,29 @@ public class ReleaseCreationInformation {
         this.pegoutCreationRskTxHash = pegoutCreationRskTxHash;
         this.pegoutBtcTx = pegoutBtcTx;
         this.pegoutConfirmationRskTxHash = pegoutConfirmationRskTxHash;
+
+        this.decodeUtxoOutpointValues(transactionReceipt);
+    }
+
+    private void decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
+        List<LogInfo> logs =  transactionReceipt.getLogInfoList();
+
+        Function pegoutTransactionCreatedEvent = BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent();
+        final byte[] pegoutTransactionCreatedSignatureTopic = pegoutTransactionCreatedEvent.encodeSignatureLong();
+
+        Optional<LogInfo> pegoutTransactionCreatedLog = logs.stream().filter((log ->
+            !log.getTopics().isEmpty() && Arrays.equals(log.getTopics().get(0).getData(),
+                pegoutTransactionCreatedSignatureTopic)
+        )).findFirst();
+
+        if(!pegoutTransactionCreatedLog.isPresent()) {
+            return;
+        }
+
+        byte[] pegoutCreatedTransactionEventData = pegoutTransactionCreatedLog.get().getData();
+        byte[] serializedOutpointValues = (byte[]) pegoutTransactionCreatedEvent.decodeEventData(pegoutCreatedTransactionEventData)[0];
+
+        this.utxoOutpointValues = Collections.unmodifiableList(UtxoUtils.decodeOutpointValues(serializedOutpointValues));
     }
 
     public Block getPegoutCreationBlock() {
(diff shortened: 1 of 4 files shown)


src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
```diff
@@ -48,6 +58,29 @@ public ReleaseCreationInformation(
         this.pegoutCreationRskTxHash = pegoutCreationRskTxHash;
         this.pegoutBtcTx = pegoutBtcTx;
         this.pegoutConfirmationRskTxHash = pegoutConfirmationRskTxHash;
+
+        this.decodeUtxoOutpointValues(transactionReceipt);
+    }
+
+    private void decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
```
apancorb, line 65: What about making this method return `utxoOutpointValues` so we are not setting instance variables as side effects. Something like:

```suggestion
        this.utxoOupointValues = decodeUtxoOutpointValues(transactionReceipt);
    }

    private List<Coin> decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
```

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
```diff
@@ -1,16 +1,26 @@
...
 import org.ethereum.core.TransactionReceipt;
+import org.ethereum.vm.LogInfo;
 
 public class ReleaseCreationInformation {
     private final Block pegoutCreationBlock;
     private final TransactionReceipt transactionReceipt;
     private final Keccak256 pegoutCreationRskTxHash;
     private final BtcTransaction pegoutBtcTx;
     private final Keccak256 pegoutConfirmationRskTxHash;
+    private List<Coin> utxoOutpointValues = Collections.emptyList();
```
apancorb, line 23: This can be made `final` if adopted what I proposed before

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
```diff
@@ -48,6 +58,29 @@ public ReleaseCreationInformation(
...
+
+        Function pegoutTransactionCreatedEvent = BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent();
+        final byte[] pegoutTransactionCreatedSignatureTopic = pegoutTransactionCreatedEvent.encodeSignatureLong();
+
+        Optional<LogInfo> pegoutTransactionCreatedLog = logs.stream().filter((log ->
+            !log.getTopics().isEmpty() && Arrays.equals(log.getTopics().get(0).getData(),
+                pegoutTransactionCreatedSignatureTopic)
+        )).findFirst();
+
+        if(!pegoutTransactionCreatedLog.isPresent()) {
```
apancorb, line 76: nit
```suggestion
        if (!pegoutTransactionCreatedLog.isPresent()) {
```

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
```diff
@@ -48,6 +58,29 @@ public ReleaseCreationInformation(
...
+    private void decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
+        List<LogInfo> logs =  transactionReceipt.getLogInfoList();
+
+        Function pegoutTransactionCreatedEvent = BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent();
+        final byte[] pegoutTransactionCreatedSignatureTopic = pegoutTransactionCreatedEvent.encodeSignatureLong();
+
+        Optional<LogInfo> pegoutTransactionCreatedLog = logs.stream().filter((log ->
+            !log.getTopics().isEmpty() && Arrays.equals(log.getTopics().get(0).getData(),
+                pegoutTransactionCreatedSignatureTopic)
+        )).findFirst();
```
apancorb, line 74: No need for the extra parenthesis
```suggestion
        Optional<LogInfo> pegoutTransactionCreatedLog = logs.stream().filter(log ->
            !log.getTopics().isEmpty() && Arrays.equals(log.getTopics().get(0).getData(),
                pegoutTransactionCreatedSignatureTopic)
        ).findFirst();
```

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
```diff
@@ -48,6 +58,29 @@ public ReleaseCreationInformation(
...
+
+        if(!pegoutTransactionCreatedLog.isPresent()) {
+            return;
+        }
+
+        byte[] pegoutCreatedTransactionEventData = pegoutTransactionCreatedLog.get().getData();
+        byte[] serializedOutpointValues = (byte[]) pegoutTransactionCreatedEvent.decodeEventData(pegoutCreatedTransactionEventData)[0];
+
+        this.utxoOutpointValues = Collections.unmodifiableList(UtxoUtils.decodeOutpointValues(serializedOutpointValues));
```
apancorb, line 83: I was thinking about this and since we are using the Optional API we can make this more functional. Something like this, wdyt?

```java
private void decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
   return transactionReceipt.getLogInfoList().stream()
        .filter(this::isPegoutTransactionCreatedLog)
        .findFirst()
        .map(LogInfo::getData)
        .map(this::decodePegoutTransactionEventData)
        .map(UtxoUtils::decodeOutpointValues)
        .map(Collections::unmodifiableList)
        .orElse(Collections.emptyList());
}

private boolean isPegoutTransactionCreatedLog(LogInfo log) {
    Function pegoutTransactionCreatedEvent = BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent();
    byte[] pegoutTransactionCreatedSignatureTopic = pegoutTransactionCreatedEvent.encodeSignatureLong();

    return !log.getTopics().isEmpty() && 
           Arrays.equals(log.getTopics().get(0).getData(), pegoutTransactionCreatedSignatureTopic);
}

private byte[] decodePegoutTransactionEventData(byte[] eventData) {
    return (byte[]) pegoutTransactionCreatedEvent.decodeEventData(eventData)[0];
}
```

apancorb (commented):
Looks very nice, added some structuring comments

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
nathanieliov: Nice idea.

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
nathanieliov: Awesome idea.

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
nathanieliov: I very like your proposal.

nathanieliov: Done

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
nathanieliov: Solved.

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
nathanieliov: Solved

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
nathanieliov: Done

src/test/java/co/rsk/federate/bitcoin/BitcoinTestUtils.java
```diff
@@ -0,0 +1,28 @@
+package co.rsk.federate.bitcoin;
+
+import co.rsk.bitcoinj.core.Coin;
+import co.rsk.bitcoinj.core.Sha256Hash;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class BitcoinTestUtils {
```
wilmerrootstock, line 9: we might improve this utility class by putting the `final` keyword at level class.

nathanieliov: Done

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
```diff
@@ -48,6 +57,31 @@ public ReleaseCreationInformation(
         this.pegoutCreationRskTxHash = pegoutCreationRskTxHash;
         this.pegoutBtcTx = pegoutBtcTx;
         this.pegoutConfirmationRskTxHash = pegoutConfirmationRskTxHash;
+
+        this.utxoOutpointValues = this.decodeUtxoOutpointValues(transactionReceipt);
+    }
+
+    private List<Coin> decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
+        return transactionReceipt.getLogInfoList().stream()
```
jeremy-then, line 65: Nice 👍🏽

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
```diff
@@ -48,6 +57,31 @@ public ReleaseCreationInformation(
         this.pegoutCreationRskTxHash = pegoutCreationRskTxHash;
         this.pegoutBtcTx = pegoutBtcTx;
         this.pegoutConfirmationRskTxHash = pegoutConfirmationRskTxHash;
+
+        this.utxoOutpointValues = this.decodeUtxoOutpointValues(transactionReceipt);
```
apancorb, line 61: nit: no need to use this leads to redundancy
```suggestion
        this.utxoOutpointValues = decodeUtxoOutpointValues(transactionReceipt);
```

apancorb (approved):
LGTM

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
```diff
@@ -348,4 +390,78 @@ void getTxInfoToSign_whenTransactionReceiptNotFoundInSubsequentBlock_shouldThrow
...
     }
+
+    @ParameterizedTest
+    @MethodSource("getTxInfoToSignArgProvider")
+    void getTxInfoToSign_whenTransactionReceiptHasPegoutTransactionCreatedEvent_returnOk(
+        int version,
+        byte[] serializedOutpointValues,
+        List<Coin> expectedOutpointValues
+    )
+        throws HSMReleaseCreationInformationException {
```
apancorb, line 401: nit: make sure we are formatting all newly introduced code

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
nathanieliov: Done

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
nathanieliov: Done

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
```diff
@@ -48,6 +58,32 @@ public ReleaseCreationInformation(
         this.pegoutCreationRskTxHash = pegoutCreationRskTxHash;
         this.pegoutBtcTx = pegoutBtcTx;
         this.pegoutConfirmationRskTxHash = pegoutConfirmationRskTxHash;
+
+        this.utxoOutpointValues = decodeUtxoOutpointValues(transactionReceipt);
+    }
+
+    private List<Coin> decodeUtxoOutpointValues(TransactionReceipt transactionReceipt) {
```
marcos-iov, line 65: Do you need to pass the TransactionReceipt as param? It's already part of the class

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
```diff
@@ -48,6 +58,32 @@ public ReleaseCreationInformation(
...
+            .findFirst()
+            .map(LogInfo::getData)
+            .map(this::decodePegoutTransactionEventData)
+            .map(UtxoUtils::decodeOutpointValues)
+            .orElse(Collections.emptyList());
+    }
+
+    private boolean isPegoutTransactionCreatedLog(LogInfo log) {
+        Function pegoutTransactionCreatedEvent = BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent();
+        final byte[] pegoutTransactionCreatedSignatureTopic = pegoutTransactionCreatedEvent.encodeSignatureLong();
```
marcos-iov, line 77: What's the difference between `encodeSignatureLong` and `encodeSignature`?

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
```diff
@@ -48,6 +58,32 @@ public ReleaseCreationInformation(
...
+
+        return !log.getTopics().isEmpty() && Arrays.equals(log.getTopics().get(0).getData(),
```
marcos-iov, line 79: Why assume it's always on index 0? There are other logs in the receipt, right? And there could be more in the future

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
```diff
@@ -348,4 +390,78 @@ void getTxInfoToSign_whenTransactionReceiptNotFoundInSubsequentBlock_shouldThrow
...
+        topics.add(DataWord.valueOf(releaseRequestedSignatureTopic));
+        topics.add(DataWord.valueOf(pegoutCreationRskTx.getHash().getBytes()));
+        topics.add(DataWord.valueOf(pegoutBtcTx.getHash().getBytes()));
+        logs.add(new LogInfo(PrecompiledContracts.BRIDGE_ADDR.getBytes(), topics, null));
+
+        CallTransaction.Function pegoutTransactionCreatedEvent = BridgeEvents.PEGOUT_TRANSACTION_CREATED.getEvent();
+        byte[] pegoutTransactionCreatedSignatureTopic = pegoutTransactionCreatedEvent.encodeSignatureLong();
+        List<DataWord> pegoutTransactionCreatedTopics = new ArrayList<>();
+        pegoutTransactionCreatedTopics.add(DataWord.valueOf(pegoutTransactionCreatedSignatureTopic));
+        pegoutTransactionCreatedTopics.add(DataWord.valueOf(pegoutBtcTx.getHash().getBytes()));
```
jeremy-then, line 417: No big deal maybe, but the real logs at the moment have the following events, in order: `update_collections`, `release_requested ` and `batch_pegout_created`,
When RSKIP428 is active, the list will be, in order:  `update_collections`, `release_requested ` and `batch_pegout_created`, `pegout_transaction_created`.

To be more realistic, do you think we should include all these events in the `logs` list? Maybe have a utility method to create the `logs` with the original events and add the `pegout_transaction_created` event in this test so it is more clear and focused?

I'm thinking about this because the actual code will be handling more events than just these two.

What do you guys think? @nathanieliov , @marcos-iov , @wilmerrootstock , @julia-zack , @apancorb .

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
```diff
@@ -348,4 +390,78 @@ void getTxInfoToSign_whenTransactionReceiptNotFoundInSubsequentBlock_shouldThrow
...
+        );
+
+        // assert
+        assertEquals(pegoutCreationBlock, pegoutCreationInfo.getPegoutCreationBlock());
+        assertEquals(pegoutCreationRskTxReceipt, pegoutCreationInfo.getTransactionReceipt());
+        assertEquals(pegoutCreationRskTx.getHash(), pegoutCreationInfo.getPegoutCreationRskTxHash());
+        assertEquals(pegoutBtcTx, pegoutCreationInfo.getPegoutBtcTx());
+        assertArrayEquals(expectedOutpointValues.toArray(), pegoutCreationInfo.getUtxoOutpointValues().toArray());
+    }
+
```
jeremy-then, line 451: What do you think about adding a test for when the `transactionReceipt` doesn't have the `pegout_transaction_created` the `utxoOutpointValues` list should be empty? Or update an existing test to include that assertion.
Let me know.

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
jeremy-then: Good point!

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
nathanieliov: It seems valid to me.  We should add them to make the test as realistic as possible.

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
jeremy-then: `encodeSignature()` returns the first 4 bytes of the encoded signature (Keccak256 hash of the signature). `encodeSignatureLong()` returns the full 32 bytes.

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
nathanieliov: There is already a test for that. Check the existing ok test.

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
jeremy-then: This is not for the logs, this is for the topics. The encoded signature (Keccak256 hash of the signature) will always be at index 0. The other indices for the same topic will be the `indexed` data of the event.

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
nathanieliov: `encodeSignatureLong` is the full signature, while `encodeSignature` is the shorter one. As @jeremy-then says.

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
jeremy-then: To visualize it better, it looks like this:

logs = [
    LogInfo {
       topics: [
           [Keccak256 hash of the event signature]: hash,                   // update_collections(address)
       ]
    },
    LogInfo {
       topics: [
           [Keccak256 hash of the event signature]: hash,                  // release_requested(bytes32,bytes32,uint256)
            [First indexed event param data]: encoded_data,
            [Second indexed event param data]: encoded_data,
       ]
    },
   LogInfo {
       topics: [
           [Keccak256 hash of the event signature]: hash,                  // batch_pegout_created(bytes32,bytes)
            [First indexed event param data]: encoded_data,
       ]
    },
    LogInfo {
       topics: [
           [Keccak256 hash of the event signature]: hash,                // pegout_transaction_created(bytes32,bytes)
            [First indexed event param data]: encoded_data,
       ]
    }
}

nathanieliov: This zero index checks topics, not logs. We assume that the zero index topic always has the encoded signature of the event, so we check that the value matches the `pegoutTransactionCreatedEvent` signature to confirm this log belongs to `pegoutTransactionCreatedEvent`

nathanieliov: According to the API of logs and events, the zero index topic always has the signature of the event, so in fact, we are not assuming, we are just following the API.

apancorb: https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP146.md#topics

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
nathanieliov: By the way, not all pegout creation rsk tx(the one with release_requested) always has a `batch_pegout_created` event; for example, migrations and rejected pegins do not include that log.

marcos-iov: Sounds like different scenarios to test

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
marcos-iov: Gotcha! Thanks

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
marcos-iov: True, my bad. Thanks!

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
nathanieliov: Done

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
nathanieliov: Done

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
```diff
@@ -1,29 +1,71 @@
...
+        Sha256Hash pegoutBtcTxHash = BitcoinTestUtils.createHash(1);
+        pegoutBtcTx = mock(BtcTransaction.class);
+        when(pegoutBtcTx.getHash()).thenReturn(pegoutBtcTxHash);
+
+        Keccak256 pegoutCreationRskTxHash = TestUtils.createHash(2);
+        pegoutCreationRskTx = mock(Transaction.class);
+        when(pegoutCreationRskTx.getHash()).thenReturn(pegoutCreationRskTxHash);
+        when(pegoutCreationRskTx.getReceiveAddress()).thenReturn(PrecompiledContracts.BRIDGE_ADDR);
+
+        Keccak256 pegoutCreationBlockHash = TestUtils.createHash(3);
```
marcos-iov, line 73: You can achieve this using BlockHeaderBuilder to create actual Block objects instead of mocks. Check https://github.com/rsksmart/powpeg-node/blob/master/src/test/java/co/rsk/federate/signing/hsm/advanceblockchain/HsmBookkeepingClientImplTest.java#L636 

Just an idea

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
```diff
@@ -348,4 +390,78 @@ void getTxInfoToSign_whenTransactionReceiptNotFoundInSubsequentBlock_shouldThrow
...
+        // Arrange
+        List<LogInfo> logs = new ArrayList<>();
+
+        CallTransaction.Function releaseRequestedEvent = BridgeEvents.RELEASE_REQUESTED.getEvent();
+        byte[] releaseRequestedSignatureTopic = releaseRequestedEvent.encodeSignatureLong();
+        List<DataWord> topics = new ArrayList<>();
+        topics.add(DataWord.valueOf(releaseRequestedSignatureTopic));
+        topics.add(DataWord.valueOf(pegoutCreationRskTx.getHash().getBytes()));
+        topics.add(DataWord.valueOf(pegoutBtcTx.getHash().getBytes()));
+        logs.add(new LogInfo(PrecompiledContracts.BRIDGE_ADDR.getBytes(), topics, null));
```
marcos-iov: Should we also add the non-indexed data? To make it as realistic as possible. We could have a utils class to create events given some data.

```
class EventsTestUtils {

     LogInfo createReleaseRequestedEvent(pegoutCreationRskTx, pegoutBtcTx, amount() { ..
```
what do you think?

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
nathanieliov: Dome

src/main/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformation.java
nathanieliov: Fixed

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
nathanieliov: Done

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
nathanieliov: Agree. I pushed the changes with `TestUtil` class. Thanks.

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
jeremy-then: 👍🏽

jeremy-then (approved):
LGMT.

src/test/java/co/rsk/federate/signing/hsm/message/ReleaseCreationInformationGetterTest.java
```diff
@@ -323,29 +427,214 @@ void getTxInfoToSign_whenTransactionReceiptNotFoundInSubsequentBlock_shouldThrow
...
+    ) throws HSMReleaseCreationInformationException {
+        // Arrange
+        List<LogInfo> logs = new ArrayList<>();
+
+        ECKey senderKey = new ECKey();
+        RskAddress senderAddress = new RskAddress(senderKey.getAddress());
+        LogInfo updateCollectionsLog = createUpdateCollectionsLog(senderAddress);
+        logs.add(updateCollectionsLog);
+
+        Coin pegoutAmount = pegoutBtcTx.getInputSum();
```
marcos-iov, line 475: Not so sure about this. The sum of all the inputs does not equal the pegout amount. There will be change

nathanieliov: You are right. I will set it equal to the minimum `bridgeConstants.getMinimumPegoutTxValue()`. A batch pegout containing a single pegout request sends the minimum. Wdyt?

nathanieliov: Done

marcos-iov (approved):

PR #274, opened 2024-06-17
Add getMessageToSign method to PowHSMSignerMessage class 

nathanieliov:
Implement getMessageToSign method. It should be able to build the message for each different version.

## Description

- Add HSM Fields
- Implement getMessageToSign method responsible for creating and populating the object message to sign by the PowHSM.
- Add private encoder outpoint values method responsible to convert the values into the format expected by the HSM
- Add outpoint values field to
- Add tests for PowHSMSignerMessageBuilder class.
- Do some refactors on ReleaseCreationInformationGetterTest and PowHSMSignerMessageTest

## Motivation and Context

To add support to HSM segwit to the powpeg node

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:
- [x] My change requires a change to the documentation.


src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
```diff
@@ -77,6 +115,59 @@ public Sha256Hash getSigHash() {
...
+    public JsonNode getMessageToSign(int version) {
+        ObjectNode messageToSend = new ObjectMapper().createObjectNode();
+        messageToSend.put(TX.getFieldName(), getBtcTransactionSerialized());
+        messageToSend.put(INPUT.getFieldName(), inputIndex);
+        if (version == HSM_VERSION_5) {
+            if (hasWitness()) {
+                populateWithSegwitValues(messageToSend);
+            } else {
+                populateWithLegacyValues(messageToSend);
+            }
```
marcos-iov, line 127: Maybe move this into its own function?

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
```diff
@@ -77,6 +115,59 @@ public Sha256Hash getSigHash() {
...
+        }
+        return messageToSend;
+    }
+
+    private JsonNode populateWithLegacyValues(ObjectNode messageToSend) {
+        messageToSend.put(SIGHASH_COMPUTATION_MODE.getFieldName(), SIGHASH_LEGACY_MODE);
+        return messageToSend;
```
marcos-iov, line 134: This `returns` but nothing is done with the returned value

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
```diff
@@ -18,33 +18,71 @@
...
+import java.util.Optional;
 import org.ethereum.core.TransactionReceipt;
 import org.spongycastle.util.encoders.Hex;
 
 public class PowHSMSignerMessage extends SignerMessage {
+
+    public static final int OUTPOINT_VALUE_LE_FORMAT_SIZE_IN_BYTES = 8;
+    public static final int HSM_VERSION_5 = 5;
+    public static final String SIGHASH_LEGACY_MODE = "legacy";
+    public static final String SIGHASH_SEGWIT_MODE = "segwit";
```
marcos-iov, line 49: why public?

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
marcos-iov: Remember to check intellij warnings, it highlights these things

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessageBuilder.java
```diff
@@ -59,7 +62,8 @@ public SignerMessage buildMessageForIndex(int inputIndex) throws SignerMessageBu
                 inputIndex,
                 txReceipt,
                 receiptMerkleProof,
-                sigHash
+                sigHash,
+                outpointValues
```
marcos-iov, line 66: If we are always passing a list of outpoint values, then is the constructor of PowHSMMessage that does not receive the outppint values used anywhere?

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
nathanieliov:
```
public JsonNode getMessageToSign(int version) {
        ObjectNode messageToSend = new ObjectMapper().createObjectNode();
        messageToSend.put(TX.getFieldName(), getBtcTransactionSerialized());
        messageToSend.put(INPUT.getFieldName(), inputIndex);
        if (version == HSM_VERSION_5) {
            populateWithHsmVersion5Values(messageToSend);
        }
        return messageToSend;
    }

    private void populateWithHsmVersion5Values(ObjectNode messageToSend) {
        if (hasWitness()) {
            populateWithSegwitValues(messageToSend);
        } else {
            populateWithLegacyValues(messageToSend);
        }
    }
``` 

Do you mean like this?

There is no prob for me in doing this change, but I don't see much benefit. Also, we agreed that soon, this is supposed to be refactored in benefit of use polymorphism. So I think it is better to focus the next effort on that way.

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessageBuilder.java
nathanieliov: It is not used anymore in the `main/java` but in some tests in  `test/java`. Thinking of keeping the PR small I just thought getting rid of the use of that constructor could be done in another task. Do you think is ok to do it in this PR?

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
nathanieliov: Just to be able to use it in the test assertions. But it's ok, I agree that better to duplicate them there.

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
nathanieliov: Solved.

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
nathanieliov: Updated.

src/test/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessageTest.java
```diff
@@ -78,109 +138,187 @@ void equality() {
...
+        }
+
+        BtcTransaction segwitPegoutBtcTx = new BtcTransaction(btcMainnetParams);
+
+        TransactionWitness txWitness = new TransactionWitness(outpointValues.size());
+        for (int i = 0; i < outpointValues.size(); i++) {
+            Coin outpointValue = outpointValues.get(i);
+
+            segwitPegoutBtcTx.addInput(fundingTransaction.getOutput(i));
+            txWitness.setPush(i, new byte[]{0x1});
```
marcos-iov, line 269: This isn't very realistic, the witness should have the same data as the script sig of a legacy transaction. I think it's important we use realistic transactions to test this.

We can consider creating a BtcTransactionBuilder to help with this if there isn't something already existent

src/test/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessageTest.java
```diff
@@ -78,109 +138,187 @@ void equality() {
...
-        assertEquals(sigHash, ((PowHSMSignerMessage) message).getSigHash());
+        // act
+        JsonNode actualMessageToSign = powHSMSignerMessage.getMessageToSign(version);
+
+        // assertions
+        Assertions.assertEquals(expectedMessageToSend, actualMessageToSign);
+    }
+
+    @Test
+    void getMessageToSign_whenSighashSegwitMode_ok() {
```
marcos-iov, line 247: Maybe we should have a test where we sign a different input than just the first one?

Or a more complete test where we create the message to sign for more than one input

nathanieliov: Actually, this test is signing a pegout with three inputs.

src/test/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessageTest.java
```diff
@@ -78,109 +138,187 @@ void equality() {
...
 
     @Test
     void getReceiptMerkleProof() {
-        String rawTx = "020000000001017001d967a340069c0b169fcbeb9cb6e0d78a27c94a41acbce762abc695aefab10000000017160014c" +
-                "fa63de9979e2a8005e6cb516b86202860ff3971ffffffff0200c2eb0b0000000017a914291a7ddc558810708149a731f39cd3c3" +
-                "a8782cfd870896e1110000000017a91425a2e67511a0207c4387ce8d3eeef498a4782e64870247304402207e0615f440bbc5035" +
-                "1fb5d8839b3fae6c74f652c9ffc9291008f4ea39f9565980220354c734511a0560367b300eecb1a7472317a995462622e06ee91" +
-                "cbe0517c17e1012102e87cd90f3cb0d64eeba797fbb8f8ceaadc09e0128afbaefb0ee9535875ea395400000000";
-        BtcTransaction btcTx = new BtcTransaction(RegTestParams.get(), Hex.decode(rawTx));
+        BtcTransaction btcTx = new BtcTransaction(RegTestParams.get(), Hex.decode(simpleRawBtcTx));
```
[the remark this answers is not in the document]
nathanieliov: Sure. This was just a quick refactor for this test, but it needs more improvement. The thing is, should we refactor those legacy tests in this PR?

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessageBuilder.java
nathanieliov: Done.

src/test/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessageTest.java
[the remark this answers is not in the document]
nathanieliov: I added this change as a requirement for the task to refactor legacy tests.

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
```diff
@@ -18,33 +18,69 @@
...
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import org.ethereum.core.TransactionReceipt;
 import org.spongycastle.util.encoders.Hex;
 
 public class PowHSMSignerMessage extends SignerMessage {
+
+    private static final int HSM_VERSION_5 = 5;
```
apancorb, line 45: I see this often, it might be a good idea to have an enum of different HSM versions so we are not using ints. Wdyt, maybe for a future task?

nathanieliov: That sounds good to me. I agree that this should be moved to the enum, where we define all the HSM versions that we support. Thanks for bringing this up. I will create a task.

nathanieliov: Task created.

marcos-iov (approved):

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
```diff
@@ -77,6 +113,48 @@ public Sha256Hash getSigHash() {
...
+        messageToSend.put(OUTPOINT_VALUE.getFieldName(), outpointValue);
+
+        TransactionWitness witness = btcTransaction.getWitness(inputIndex);
+        String encodedWitnessScript = Hex.toHexString(witness.getScriptBytes());
+        messageToSend.put(WITNESS_SCRIPT.getFieldName(), encodedWitnessScript);
+    }
+
+    private long getOutpointValueForInputIndex() {
+        return Optional.of(outpointValues)
+            .filter(values -> inputIndex >= 0 && inputIndex < outpointValues.size())
```
jeremy-then, line 147: The `values` variable is not being used here. I guess the `outpointValues.size()` it's doing its job, since they are supposed to be the same, right? Since `values` is already there, what if we make use of it?

jeremy-then (approved):
LGTM.

src/main/java/co/rsk/federate/signing/hsm/message/PowHSMSignerMessage.java
nathanieliov: Good catch

nathanieliov: Done

apancorb (approved):
LGTM

PR #275, opened 2024-06-17
feat(ci): replaces circleci w/ github workflow 

jurajpiar:
Solves [CORE-5765](https://rsklabs.atlassian.net/browse/CORE-5765) by removing the CircleCI configuration and migrating it to GHA workflow one.
The sonarqube required some additions to the `build.gradle` file.

> [!IMPORTANT]
Just before merging this PR we should remove the CircleCI integration in the repo settings @rsksmart/security.

> [!WARNING]
@rsksmart/security We might want to consider shortening the default artefact storage period globally for this repo, as we only need them stored for the execution. See [retention-days input](https://github.com/actions/upload-artifact?tab=readme-ov-file#inputs) for more info. Unless you'd prefer for me to use this param locally for each artefact?


rmoreliovlabs:
LGTM!

rmoreliovlabs (approved):

build.gradle
```diff
@@ -33,6 +33,7 @@ jacocoTestReport {
     reports {
         xml.required = true
     }
+    finalizedBy jacocoTestCoverageVerification
```
jurajpiar, line 36: just added this so that coverage issues are detected earlier (presumably)

jeremy-then (dismissed):
LTGM.

fmacleal (commented):
Good job, I am sure that it went through a long way till reach this concise and straightforward version.

I have made comments, let me know your thoughts. After this I will give the approval.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,235 @@
...
+                    dependency.useTarget targetProject
+                  }
+                }
+              }
+            }
+          }
+          EOF
+          if [ -d ".gradle" ]; then rm -rfv .gradle; fi
+          ./configure.sh
+          ./gradlew --no-daemon --settings-file temp-settings.gradle dependencies
```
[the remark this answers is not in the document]
jurajpiar: personally, I don't like fragmentation of github workflow code, but should others support this, I will change it.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,235 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Download rskj
+        uses: actions/download-artifact@v4
+        with:
+          name: rskj
+          path: rskj
```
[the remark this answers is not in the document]
jurajpiar: only if we do all the work in the same step, which to me seems dirty.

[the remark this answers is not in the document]
fmacleal: It looks like, but it's part of the powpeg process build it, the rskj it's a dependent library. Doesn't make much sense having it separated from the powpeg build, IMHO. 

Because you can't build powpeg without having a proper version of the RSKj.

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
fmacleal: That's the thing, this isn't supposed to be github workflow code, it's an internal thing from powpeg that requires it to build rskj. My suggestion is exactly to separate the concerns and responsibilities and let the github workflow takes care of the workflow, and let the internal ways of build an application separated from it. Because if we change this at some point, we update the script and we don't need to change the workflow.

But again, it's a personal opinion, do as you wish. It doesn't prevent the approval.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,235 @@
...
+      - name: Check out appropriate rskj reference
+        working-directory: rskj
+        run: |
+          git switch "${{ env.CHECKOUT_REF }}"
+
+      - name: Persist RSKJ
+        uses: actions/upload-artifact@v4
+        with:
+          name: rskj
+          path: rskj
```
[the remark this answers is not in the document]
fmacleal: It's way more complex to change this here on the workflow than in a bash script, but, it's your call at the end. Do as you wish. It was more a thought, but the final decision it's yours. ;)

fmacleal (dismissed):
LGTM

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,256 @@
...
+          else
+            echo "Using master for RSKj"
+            CHECKOUT_REF="master"
+          fi
+          echo "CHECKOUT_REF=$CHECKOUT_REF" >> $GITHUB_ENV
+
+      - name: Check out appropriate rskj reference
+        working-directory: rskj
+        run: |
+          git switch "${{ env.CHECKOUT_REF }}"
```
lucasvuotto, line 79: Manage this as `POW_REF` was managed in the previous step.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,256 @@
...
+        run: |
+          IS_RSKJ_BRANCH=`git ls-remote --heads origin $PR_BRANCH`
+          if test -n "${IS_RSKJ_BRANCH}"; then
+            echo "Found matching branch name in RSKj repo"
+            CHECKOUT_REF="$PR_BRANCH"
+          else
+            echo "Using master for RSKj"
+            CHECKOUT_REF="master"
+          fi
+          echo "CHECKOUT_REF=$CHECKOUT_REF" >> $GITHUB_ENV
```
diego-jeronymo, line 58: Please, just make sure that any external user can't control the name or the value of this `GITHUB_ENV` environment variable.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,256 @@
...
+        run: |
+          IS_RSKJ_REF=`git ls-remote --refs $POW_REF`
+          if test -n "${IS_RSKJ_REF}"; then
+            echo "Found matching ref in RSKj"
+            CHECKOUT_REF="$POW_REF"
+          else
+            echo "Using master for RSKj"
+            CHECKOUT_REF="master"
+          fi
+          echo "CHECKOUT_REF=$CHECKOUT_REF" >> $GITHUB_ENV
```
diego-jeronymo, line 74: Please, just make sure that any external user can't control the name or the value of this `GITHUB_ENV` environment variable.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,256 @@
...
+
+      - name: Persist RSKJ
+        uses: actions/upload-artifact@v4
+        with:
+          name: rskj
+          path: rskj
+  
+  build_federator_node:
+    runs-on: ubuntu-latest
+    container: openjdk:8-jdk
```
Vovchyk, line 89: could we use java17 instead? like in [here](https://github.com/rsksmart/rskj/pull/2624/files#diff-48c0ee97c53013d18d6bbae44648f7fab9af2e0bf5b0dc1ca761e18ec5c478f2R14-R23)?

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,256 @@
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: build_files
+          path: |
+            ./
+            !rskj
+
+  federator-tests:
+    runs-on: ubuntu-latest
+    container: openjdk:8-jdk
```
Vovchyk, line 154: same here

Vovchyk (changes requested):

.github/workflows/build_and_test.yml
rmoreliovlabs: Thanks for highlighting this. In our GitHub Actions workflow, we ensure that all inputs and conditions influencing `GITHUB_ENV` are validated and controlled within the workflow itself. This ensures that only trusted sources influence our CI/CD processes. If you have any specific suggestions or additional security measures in mind, I'm open to discussing them further

.github/workflows/build_and_test.yml
rmoreliovlabs: The workflow  already aligns well with how `POW_REF` was handled in the previous step. Both `CHECKOUT_REF` and `POW_REF` are set based on whether it's a pull request or a direct push event. We check if the branch or reference exists in the repository using git ls-remote and default to master if there's no match found. This approach ensures consistency and clarity throughout the workflow. It makes sure we're always checking out the right branch or reference depending on how the GitHub action was triggered.

If there's anything specific you think needs adjusting or any other concerns, just let me know!

.github/workflows/build_and_test.yml
rmoreliovlabs: Addressed. Switched to Java 17

.github/workflows/build_and_test.yml
rmoreliovlabs: solved

.github/workflows/build_and_test.yml
rmoreliovlabs: addressed above ^^

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,256 @@
...
+          path: rskj
+          fetch-depth: 0
+      
+      - name: Determine branch to checkout for PR
+        if: github.event_name == 'pull_request'
+        working-directory: rskj
+        env:
+          PR_BRANCH: ${{ github.head_ref }}
+        run: |
+          IS_RSKJ_BRANCH=`git ls-remote --heads origin $PR_BRANCH`
```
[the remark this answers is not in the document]
lucasvuotto: ugh, can't believe I missed the `PR_BRANCH` definition just before it :(

I meant to quote the `$PR_BRANCH` use, not the definition. In the definition it makes no difference, because it's YAML. In its use, `git ls-remote --heads origin $PR_BRANCH` will do funny things if it `$PR_BRANCH` isn't quoted and, for example, has spaces.

Except in the case that you fully control the value of variables and you know that in all cases it doesn't include characters from the input field separators, you should quote the variable usage in shell scripts.

.github/workflows/build_and_test.yml
lucasvuotto: With the current implementation, I can create a branch called `main";env;echo"` which will render the command as

```
git switch "main";env;echo""
```

showing all the environments variable in the workflow. I can change `env` for a `curl` and further leak information.

If you replace it with

```
env:
  CHECKOUT_REF: ${{ env.CHECKOUT_REF }}
run: |
  git switch "$CHECKOUT_REF"
```

the issue is gone, because it's the shell doing the replacement and not the template engine of GHA.

lucasvuotto: Bonus comment: `CHECKOUT_REF` is defined in the environment only in the case that the event isn't a pull request (the condition of the previous step), yet this step is always ran.

jeremy-then (dismissed):
LGTM.

.github/workflows/build_and_test.yml
rmoreliovlabs: Thanks for the clarification! I've implemented the solution you proposed

nathanieliov (approved):
LGTM

fmacleal (approved):

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,268 @@
...
+      - master
+      - "*-rc"
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+    branches:
+      - "**"
+
```
bcodesido, line 15: Can you please specify the workflow permissions?
Should be 
```
permissions:
  contents: read
```

rmoreliovlabs: addressed!

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
rmoreliovlabs: Thanks for pointing that out! I've updated the workflow to quote both PR_BRANCH and POW_REF. They both look something like this:

```
env:
  PR_BRANCH: ${{ github.head_ref }}
run: |
  IS_RSKJ_BRANCH=$(git ls-remote --heads origin "$PR_BRANCH")
```
Appreciate the help in catching that!

lucasvuotto (approved):
LGTM

bcodesido (approved):
LGTM

diego-jeronymo (approved):
LGTM

marcos-iov:
pipeline:run

PR #282, opened 2024-06-28
Create enum for supported HSM Versions

nathanieliov:
Create an enum for supported HSM Versions.

## Description

- Create an enum for supported HSM Versions
- Replace HSM version literals for SignerVersion Enum

## Motivation and Context

 HSM versions are being used from different places using static class fields or method parameters. Thus, to avoid code duplication, stop using `int` literals in benefit of using an enum and to have all the supported HSM version put in a single place.

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:
- [ ] My change requires a change to the documentation.


src/main/java/co/rsk/federate/signing/SignerVersion.java
```diff
@@ -0,0 +1,20 @@
+package co.rsk.federate.signing;
+
+public enum SignerVersion {
```
apancorb, line 3: What about making the enum implement the `Comparable<SignerVersion>` interface. This way we can do conditional checks. 

```java
    @Override
    public int compareTo(SignerVersion other) {
        return Integer.compare(this.version, other.version);
    }
```

src/main/java/co/rsk/federate/FedNodeRunner.java
```diff
@@ -138,7 +138,7 @@ public void run() throws Exception {
             int hsmVersion = protocol.getVersion();
```
apancorb, line 138: The idea would be to have something like this:

```suggestion
            SignerVersion hsmVersion = SignerVersion.valueOf(protocol.getVersion());
```

Maybe to change `protocol.getVersion` to return the enum directly would be better.

apancorb (commented):
Looks good overall, but the idea is to be able to use the enums as constants not the underlying integers. By implementing the `Comparable` interface we can do the conditional checks as well.

src/main/java/co/rsk/federate/signing/SignerVersion.java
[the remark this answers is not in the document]
nathanieliov: Version 1 is a key file. This is why I named it as `SignerVersion`.

nathanieliov: This is not possible since the `enum` class already implements comparable and sets the implemented method as final. This means that the method cannot be overridden.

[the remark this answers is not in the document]
nathanieliov: I understand your point better now, and I kind of agree with you, but that refactor is more extensive than you may think. Check [this class](https://github.com/rsksmart/powpeg-node/blob/77c22f1d9c2458e53a2606e60677432ec1b6f11f/src/main/java/co/rsk/federate/signing/hsm/client/HSMSigningClientProvider.java#L47) out please. There is an `HSMSigningClientV1`, which, from my understanding, should not be named with the prefix `HSM.` since I understand `HSM` is not for `v1`.

So, I think a couple of refactors/improvement opportunities should be addressed in its specific task, but not this one because it looks out of scope to me.

Let me know what you think. Thanks.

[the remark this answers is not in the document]
nathanieliov: Done

src/main/java/co/rsk/federate/signing/SignerVersion.java
```diff
@@ -0,0 +1,20 @@
...
+    VERSION_1(1),
+    VERSION_2(2),
+    VERSION_3(3),
+    VERSION_4(4),
+    VERSION_5(5),
```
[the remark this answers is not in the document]
nathanieliov: Done

src/main/java/co/rsk/federate/FedNodeRunner.java
nathanieliov: Done

src/main/java/co/rsk/federate/signing/SignerVersion.java
```diff
@@ -0,0 +1,20 @@
...
+    ;
+
+    private int versionNumber;
```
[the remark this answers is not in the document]
nathanieliov: Done

src/main/java/co/rsk/federate/FedNodeRunner.java
apancorb: I dont think you pushed your changes to the branch

nathanieliov: Oh, sorry, you're right; I didn't add this change. The thing is, I think this change should be part of the second refactor where the use of the hsmVersion as `int` is replaced for the enum. So, I decided that it would be better to do that change altogether with the big refactor. Wdyt?

apancorb: Sounds good :)

apancorb (approved):

src/main/java/co/rsk/federate/signing/hsm/HSMVersion.java
```diff
@@ -0,0 +1,20 @@
+package co.rsk.federate.signing.hsm;
+
+public enum HSMVersion {
+    V1(1),
+    V2(2),
+    V3(3),
```
marcos-iov, line 6: We can do this in another task, but I propose we remove V3. There is no HSM v3 ever released

src/test/java/co/rsk/federate/signing/hsm/advanceblockchain/ConfirmedBlocksProviderTest.java
```diff
@@ -21,8 +22,8 @@
 
 class ConfirmedBlocksProviderTest {
 
-    private final int HSM_VERSION_2 = 2;
-    private final int HSM_VERSION_4 = 4;
+    private static final int HSM_VERSION_2 = HSMVersion.V2.getNumber();
+    private static final int HSM_VERSION_4 = HSMVersion.V4.getNumber();
```
marcos-iov, line 26: I'd remove these veriables and use the enum values directly

src/test/java/co/rsk/federate/signing/hsm/advanceblockchain/HsmBookkeepingClientImplTest.java
```diff
@@ -72,8 +73,8 @@
  * Created by Kelvin Isievwore on 14/03/2023.
  */
 class HsmBookkeepingClientImplTest {
-    private final static int VERSION_TWO = 2;
-    private final static int VERSION_FOUR = 4;
+    private static final int VERSION_TWO = HSMVersion.V2.getNumber();
+    private static final int VERSION_FOUR = HSMVersion.V4.getNumber();
```
marcos-iov, line 77: same here

src/test/java/co/rsk/federate/signing/hsm/client/PowHSMSigningClientBtcTest.java
```diff
@@ -84,7 +85,7 @@
 
 class PowHSMSigningClientBtcTest {
 
-    private static final int HSM_VERSION_2 = 2;
+    private static final int HSM_VERSION_2 = HSMVersion.V2.getNumber();
```
marcos-iov, line 88: and here

src/test/java/co/rsk/federate/signing/hsm/client/PowHSMSigningClientBtcTest.java
```diff
@@ -379,14 +380,14 @@ private static HSMSignature createMockSignature() {
     private static List<Arguments> legacyPegoutArgProvider() {
         List<Arguments> arguments = new ArrayList<>();
 
-        int[] hsmVersions = {2, 4, 5};
-        for (int hsmVersion : hsmVersions) {
+        HSMVersion[] hsmVersions = {HSMVersion.V2, HSMVersion.V4, HSMVersion.V5};
```
marcos-iov, line 383: Would it make sense to have a method in the enum that returns PowHSM versions?

src/main/java/co/rsk/federate/FedNodeRunner.java
marcos-iov: Did you create a task for this already?

src/main/java/co/rsk/federate/signing/hsm/HSMVersion.java
nathanieliov: There are lines of codes referencing hsm version 3, check [this](https://github.com/rsksmart/powpeg-node/blob/master/src/main/java/co/rsk/federate/signing/hsm/advanceblockchain/ConfirmedBlocksProvider.java#L97) and [this](https://github.com/rsksmart/powpeg-node/blob/master/src/main/java/co/rsk/federate/signing/hsm/advanceblockchain/HsmBookkeepingClientImpl.java#L169).

Should these conditionals be changed then?

marcos-iov: Yes, but in another task

nathanieliov: Task created.

src/main/java/co/rsk/federate/FedNodeRunner.java
nathanieliov: Done

src/test/java/co/rsk/federate/signing/hsm/client/PowHSMSigningClientBtcTest.java
nathanieliov: Done

src/test/java/co/rsk/federate/signing/hsm/client/PowHSMSigningClientBtcTest.java
nathanieliov: Done

src/test/java/co/rsk/federate/signing/hsm/advanceblockchain/HsmBookkeepingClientImplTest.java
nathanieliov: Done

src/test/java/co/rsk/federate/signing/hsm/advanceblockchain/ConfirmedBlocksProviderTest.java
nathanieliov: Done

src/main/java/co/rsk/federate/signing/hsm/HSMVersion.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public int getNumber() {
+        return number;
+    }
+
+    public static List<HSMVersion> getPowHSMVersions() {
+        return Arrays.asList(V2, V3, V4, V5);
+    }
+
+    public static boolean isPowHSM(int version) {
```
marcos-iov, line 28: Guess the plan is that this goes away once we do the changes to manage HSMVersion objects instead of `ints` right? We should only use the one below

Is that part of the new task definition?

src/test/java/co/rsk/federate/signing/hsm/client/PowHSMSigningClientBtcTest.java
```diff
@@ -84,7 +85,7 @@
 
 class PowHSMSigningClientBtcTest {
 
-    private static final int HSM_VERSION_2 = 2;
+    private static final HSMVersion HSM_VERSION_2 = HSMVersion.V2;
```
marcos-iov, line 88: Remove this? Use the enum directly as in other places

src/main/java/co/rsk/federate/signing/hsm/HSMVersion.java
nathanieliov: Correct, that's the plan. Added that requirement to the task.

src/test/java/co/rsk/federate/signing/hsm/client/PowHSMSigningClientBtcTest.java
nathanieliov: Good catch. I missed this one.

marcos-iov (approved):

PR #283, opened 2024-07-01
Build configuration files for tests

apancorb:
### Summary

In this PR we present a way for tests that require to use custom configuration files to be able to do so without the need of mocking them. This changes introduces the `TestConfigBuilder` class that allows to create `Config` objects in memory as if they were parsed from a real configuration file.


src/test/java/co/rsk/federate/config/TestConfigBuilder.java
```diff
@@ -0,0 +1,26 @@
+package co.rsk.federate.config;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import com.typesafe.config.ConfigValueFactory;
+
+public class TestConfigBuilder {
```
marcos-iov, line 7:
```suggestion
public class PowpegConfigBuilder {
```
What do you think?

src/test/java/co/rsk/federate/FedNodeRunnerTest.java
```diff
@@ -516,34 +515,26 @@ private SignerConfig getHSMBTCSignerConfig(int version) throws HSMClientExceptio
...
-        when(hsmConfig.getString("host")).thenReturn("127.0.0.1");
-        when(hsmConfig.hasPath("port")).thenReturn(true);
-        when(hsmConfig.getInt("port")).thenReturn(9999);
-        when(hsmConfig.getString("keyId")).thenReturn("m/44'/137'/1'/0/0");
-        return mstSignerConfig;
+        Config config = TestConfigBuilder.builder()
+            .withValue("type", "hsm")
+            .withValue("host", "127.0.0.1")
+            .withValue("port", 9999)
+            .withValue("keyId", "m/44'/137'/1'/0/0")
```
marcos-iov, line 536: Perhaps all these could be part of a single method?

`withHsmSigner(host, port, keyid)` where the `type` is automatically set to `hsm`

We could also have one with default values `withHsmSigner()` where it uses all these values by default

src/test/java/co/rsk/federate/FedNodeRunnerTest.java
```diff
@@ -443,69 +439,72 @@ void test_with_no_config() throws Exception {
...
-    private SignerConfig getMSTSignerConfig() {
-        SignerConfig mstSignerConfig = mock(SignerConfig.class);
-        when(mstSignerConfig.getId()).thenReturn("MST");
-        when(mstSignerConfig.getType()).thenReturn("keyFile");
-        when(mstSignerConfig.getConfig()).thenReturn(keyFileConfig);
-        return mstSignerConfig;
+    private SignerConfig getMSTSignerConfig(String path) {
+        Config config = TestConfigBuilder.builder()
+            .withValue("type", "keyFile")
+            .withValue("path", path)
```
marcos-iov, line 484: This could be`withKeyFileSigner(path)`

src/test/java/co/rsk/federate/FedNodeRunnerTest.java
```diff
@@ -443,69 +439,72 @@ void test_with_no_config() throws Exception {
...
+            .withValue("host", "127.0.0.1")
+            .withValue("port", 9999)
+            .withValue("keyId", "m/44'/0'/0'/0/0")
+            .withValue("socketTimeout", 20000)
+            .withValue("maxAttempts", 3)
+            .withValue("intervalBetweenAttempts", 2000)
+            .withValue("bookkeeping.informerInterval", 500000L)
+            .withValue("bookkeeping.maxAmountBlockHeaders", 1000)
+            .withValue("bookkeeping.maxChunkSizeToHsm", 100)
+            .withValue("bookkeeping.stopBookkeepingScheduler", true)
```
marcos-iov, line 457: Perhaps all these could also be grouped somehow

marcos-iov (commented):
Looking good, but I think we could make the builder a bit more helpful by adding specific methods to create certain common configs

src/test/java/co/rsk/federate/config/TestConfigBuilder.java
nathanieliov: Agree

nathanieliov (approved):
LGTM

src/test/java/co/rsk/federate/config/TestConfigBuilder.java
apancorb: Sounds good, I have renamed the class

src/test/java/co/rsk/federate/FedNodeRunnerTest.java
apancorb: Sounds good, what I have done is added the method with `host` and `port` already set except the `keyId`. The keyId changes depending from `BTC, RSK, MST` so a default here will not be possible. Wdyt with how is set up currently?

src/test/java/co/rsk/federate/FedNodeRunnerTest.java
apancorb: Sounds good, I have updated.

src/test/java/co/rsk/federate/FedNodeRunnerTest.java
apancorb: I feel like if we add them all in the `withHsmSigner` method then it could be too granular and not very flexible. wdyt?

src/test/java/co/rsk/federate/config/PowpegConfigBuilder.java
```diff
@@ -0,0 +1,41 @@
+package co.rsk.federate.config;
+
+import static com.typesafe.config.ConfigValueFactory.fromAnyRef;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+public class PowpegConfigBuilder {
```
marcos-iov, line 8: Goind through this again, looks like this builder is only building the config for different signers. Is that right?

If that's the case we should rename it, `SignerConfigBuilder` or something of the sort.

And should this return a `Config` object or the `SignerConfig` object directly? They `key` could be passed to the builder constructor or to the `build` method. What do you think?

src/test/java/co/rsk/federate/FedNodeRunnerTest.java
marcos-iov: That's true.

Perhaps a method `withBookkeepingInfo` and receive all the data as params? To group things a bit and not force the one who is using this to know the name of each field that needs to be assigned

src/test/java/co/rsk/federate/config/SignerConfigBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+    this.config = this.config.withValue("host", fromAnyRef("127.0.0.1"));
+    this.config = this.config.withValue("port", fromAnyRef(9999));
+    this.config = this.config.withValue("keyId", fromAnyRef(keyId));
+    this.config = this.config.withValue("socketTimeout", fromAnyRef(20000));
+    this.config = this.config.withValue("maxAttempts", fromAnyRef(3));
+    this.config = this.config.withValue("intervalBetweenAttempts", fromAnyRef(2000));
+    return this;
+  }
+
+  public SignerConfigBuilder withHsmBookkeepingInfo() {
```
marcos-iov, line 36:
```suggestion
  public SignerConfigBuilder withHsmBookkeepingInfo(BigInteger difficultyTarget, long informerInterval ....) {
```
This is what I had in mind actually.

That way it can called simply `withHsmBookkeepingInfo(target, interval, ...` and there's no need to know the name of the fields that need to be set.

Makes sense? 

If you think it makes sense to also have some default values, perhaps we could set those when calling `withHsmSigner`. We can have a quick chat and settle on this

src/test/java/co/rsk/federate/FedNodeRunnerTest.java
```diff
@@ -443,69 +439,49 @@ void test_with_no_config() throws Exception {
...
-    private SignerConfig getMSTSignerConfig() {
-        SignerConfig mstSignerConfig = mock(SignerConfig.class);
-        when(mstSignerConfig.getId()).thenReturn("MST");
-        when(mstSignerConfig.getType()).thenReturn("keyFile");
-        when(mstSignerConfig.getConfig()).thenReturn(keyFileConfig);
-        return mstSignerConfig;
+    private SignerConfig getMSTSignerConfig(String path) {
+        return SignerConfigBuilder.builder()
+            .withKeyFileSigner(path)
+            .build("MST");
```
marcos-iov, line 473: We could use PowPegNodeKeyId for these keys instead of the literals

src/test/java/co/rsk/federate/config/SignerConfigBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+public class SignerConfigBuilder {
+  private Config config;
+
+  private SignerConfigBuilder() {
+    this.config = ConfigFactory.empty();
+  }
+
+  public static SignerConfigBuilder builder() {
+    return new SignerConfigBuilder();
+  }
```
marcos-iov, line 18: Out of curiosity, any advantage of doing this instead of a public constructor?

apancorb: Not an advantage but is more related to readability and clarity for me. I also used extensively [lombok](https://projectlombok.org/features/Builder) and this is how it implements the builder pattern.

src/test/java/co/rsk/federate/FedNodeRunnerTest.java
apancorb: Sounds good, I have updated it

src/test/java/co/rsk/federate/config/SignerConfigBuilder.java
apancorb: Okay I see, I have updated the method with this approach. I also allowed it for the `difficultyTarget` to be nullable so we can use it for the test case where `difficultyTarget` is not set in the config. Let me know if that is fine.

marcos-iov (approved):

PR #307, opened 2024-08-05
Fix refs

nagarev:
As seen in the screenshots, there's no difference in using `refs/heads/master` or `master` for this command:

Screenshot 2024-08-05 at 13 55 24

Screenshot 2024-08-05 at 13 55 50

so I'm rolling back the variable `POW_REF` to it's previous value `github.ref` instead of `github.ref_name`.


marcos-iov:
pipeline:run

marcos-iov (approved):

PR #310, opened 2024-09-04
Adding Rootstock Integration Tests workflow to Powpeg node repo

rmoreliovlabs:


.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+          POWPEG_BRANCH="master"
+          RIT_BRANCH="${{ github.event.inputs.rit-branch || 'main' }}"
+          RSKJ_BRANCH="${{ github.event.inputs.rskj-branch || 'master' }}"
+
+          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            POWPEG_BRANCH="${{ github.event.pull_request.head.ref }}"
+          elif [[ "${{ github.event_name }}" == "push" ]]; then
+            if [[ "${{ github.ref }}" == refs/heads/*-rc ]]; then
+              POWPEG_BRANCH="${{ github.ref }}"
+            fi
```
fmacleal, line 40: question:

This condition "${{ github.ref }}" == refs/heads/*-rc can be added as an AND in the previous if couldn't it? If yes, it will make it simpler, I suggest to add it there. Also, I noted that maybe it's missing a double quote.

```suggestion
          elif [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/*-rc" ]]; then
              POWPEG_BRANCH="${{ github.ref }}"
```

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+          repository: rsksmart/powpeg-node
+          ref: ${{ env.POWPEG_BRANCH }}
+          fetch-depth: 1
+
+      - name: Checkout Rootstock Integration Tests Repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 #v4.1.7
+        with:
+          repository: rsksmart/rootstock-integration-tests
+          ref: ${{ env.RIT_BRANCH }}
+          fetch-depth: 0
```
fmacleal, line 62: suggestion:

I don't think this part is necessary, it won't impact which RIT or Powpeg branch will be used to execute the tests at the end. Because the container-action clones all the repos used and does the checkout again on the branch passed as parameter, [as you can see here](https://github.com/rsksmart/rootstock-integration-tests/blob/main/container-action/entrypoint.sh#L18-L40). Only thing is that you are missing to pass `RIT_BRANCH` on the next step. I suggest remove these steps and pass the RIT_BRANCH as input to the action.

```suggestion
```

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Run Rootstock Integration Tests
+        uses: rsksmart/rootstock-integration-tests@497172fd38dcfaf48c77f9bb1eeb6617eef5eed6 #v1
+        with:
+          rskj-branch: ${{ env.RSKJ_BRANCH }}
+          powpeg-node-branch: ${{ env.POWPEG_BRANCH }}
```
fmacleal, line 68: suggestion:

As commented above, you need to pass the RIT branch here. The container-action is ready to receive it, as you can see [here](https://github.com/rsksmart/rootstock-integration-tests/blob/main/action.yml#L14). 

```suggestion
          powpeg-node-branch: ${{ env.POWPEG_BRANCH }}
          rit-branch: ${{ env. RIT_BRANCH }}

```

.github/workflows/rit.yml
```diff
@@ -0,0 +1,68 @@
+name: Rootstock Integration Tests
+
+on:
+  pull_request:
+    types: [opened, reopened]
+    branches:
+      - "**"
```
fmacleal, line 7: suggestion:

Probably makes sense consider what @bcodesido said in the other [PR](https://github.com/rsksmart/rskj/pull/2701/files#r1745501462).

fmacleal (changes requested):
Great job, I think we are almost there and there a few things to change and we are done. 😃

.github/workflows/rit.yml
rmoreliovlabs: Addressed!
```
if [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/*-rc" ]]; then
            POWPEG_BRANCH="${{ github.ref }}"
          elif [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
            POWPEG_BRANCH="${{ github.ref_name }}"
          fi
```

.github/workflows/rit.yml
rmoreliovlabs: Addressed! Removed them

.github/workflows/rit.yml
rmoreliovlabs: Addressed, removed the pull request event

.github/workflows/rit.yml
rmoreliovlabs: Added!

.github/workflows/rit.yml
```diff
@@ -0,0 +1,46 @@
+name: Rootstock Integration Tests
+
+on:
```
fmacleal, line 3: question:

I think the issue with the pull_request trigger was the source branch that was set to all with **. Don't we want to run it at least once for a PR opened inside our organization? Maybe add an if on the job rootstock-integration-tests with something like:

    if: github.event.pull_request.head.repo.full_name == 'rsksmart/rskj'

Buuut, I don't see it as a big deal, because we can always run manually and we were already doing that triggering with a comment. So, to me, the way that it's set now it's very good and more than enough. 😃

fmacleal (dismissed):
Good job! I just have a question regarding your last change, but it doesn't prevent the approval. I think this way it will fill the initial requirements we have for this pipeline, we can always change if needed. :)

.github/workflows/rit.yml
rmoreliovlabs: nice suggestion! adding a condition for PRs from `rsksmart/rskj` is a great idea for the future. For now, I think we should stick with manual and push event triggers to keep things simple and efficient. We can revisit this if we see a need for more automated PR testing

fmacleal (approved):

bcodesido (approved):
LGTM

.github/workflows/rit.yml
```diff
@@ -0,0 +1,51 @@
...
+          elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
+            POWPEG_BRANCH="${{ github.head_ref }}"
+          fi
+
+          echo "RSKJ_BRANCH=$RSKJ_BRANCH" >> $GITHUB_ENV
+          echo "RIT_BRANCH=$RIT_BRANCH" >> $GITHUB_ENV
+          echo "POWPEG_BRANCH=$POWPEG_BRANCH" >> $GITHUB_ENV
+
+      - name: Run Rootstock Integration Tests
+        uses: rsksmart/rootstock-integration-tests@497172fd38dcfaf48c77f9bb1eeb6617eef5eed6 #v1
```
marcos-iov, line 47: Out of curiosity, what's this? A commit hash?

.github/workflows/rit.yml
```diff
@@ -0,0 +1,51 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - name: Set Branch Variables
+        id: set-branch-variables
+        run: |
+          # Default values
+          POWPEG_BRANCH="master"
+          RIT_BRANCH="${{ github.event.inputs.rit-branch || 'main' }}"
+          RSKJ_BRANCH="${{ github.event.inputs.rskj-branch || 'master' }}"
```
marcos-iov, line 32: Is there a way to somehow re-use the default branch values that are defined above?

.github/workflows/rit.yml
bcodesido: Indeed. It's a way to pin a dependency. By using the commit hash, we are explicitly setting a "version" instead of using a mutable version or range of versions.

More about can be found [here](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies)

.github/workflows/rit.yml
fmacleal: Hello, yes, it seems possible. Do you think it's better to re-use it? Because anyway, to change these default values we would have to change the wofklow yaml anyway.

fmacleal: Hello, yes, it seems possible. Do you think it's better to re-use it? Nevertheless, to change these default values we would have to change the worfklow yaml anyway.


We would have to do something like this:


```suggestion
          RIT_BRANCH="${{ github.event.inputs.rit-branch || github.event.inputs['rit-branch'].default }}"
          RSKJ_BRANCH="${{ github.event.inputs.rskj-branch || github.event.inputs['rskj-branch'].default }}"
```

fmacleal: Ops, the suggestion failed. Probably because the inputs are define only if the pipeline is triggered via workflow_dispatch. This logic wouldn't work for the commits in PR.

Let's let as it was? Otherwise we will need an even bigger if like:

`RSKJ_BRANCH="${{ github.event.inputs.rskj-branch || github.event.inputs['rskj-branch'].default || 'main' }}"`

marcos-iov: No big deal, I wouldn't expect the default branches to change anytime soon. Just the good practice of avoid repetition (DRY). But in this case I'd go with the simplest solution

PR #312, opened 2024-09-16
Fixing some issues on RIT workflow

fmacleal:
Issues fixed:
1 - Remove the trigger on push for master and *-rc -> we don't need that anymore since we are running it for every commit on PRs against these branches. 2 - Send the status of the execution to slack
3 - Solve the issue that the status of an execution triggered via workflow doesn't update the PR status. And if we need to use a different branch for RIT, Powpeg and RIT in the PR, it won't change the status to green after the workflow_dispatch execution -> We will do that adding a parsing on the description, looking for the branch that need to be used for that PR. If there is no branch configured on the description, we will try to use the default one.

Branches to be used by RIT:
`fed:vovchyk/jackson-ver-bump`
`rskj:vovchyk/jackson-ver-bump`

### DIFF
--- a/.github/workflows/rit.yml
+++ b/.github/workflows/rit.yml
@@ -1,8 +1,6 @@
 name: Rootstock Integration Tests
 
 on:
-  push:
-    branches: ["master", "*-rc"]
   pull_request:
     types: [ opened, synchronize, reopened ]
     branches: [ "master", "*-rc" ]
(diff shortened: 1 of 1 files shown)


Vovchyk (changes requested):
before we were expecting branch names inside back apostrophes, eg., \`fed:vovchyk/jackson-ver-bump\`. Could we continue doing it?

.github/workflows/rit.yml
```diff
@@ -23,29 +21,125 @@ jobs:
...
+          echo "BUILD_URL=$BUILD_URL" >> $GITHUB_ENV
 
       - name: Run Rootstock Integration Tests
         uses: rsksmart/rootstock-integration-tests@497172fd38dcfaf48c77f9bb1eeb6617eef5eed6 #v1
         with:
           rskj-branch: ${{ env.RSKJ_BRANCH }}
           powpeg-node-branch: ${{ env.POWPEG_BRANCH }}
-          rit-branch: ${{ env. RIT_BRANCH }}
\ No newline at end of file
+          rit-branch: ${{ env. RIT_BRANCH }}
```
lucasvuotto, line 111:
```suggestion
          rit-branch: ${{ env.RIT_BRANCH }}
```

lucasvuotto (changes requested):

marcos-iov (approved):

Vovchyk (approved):

marcos-iov:
pipeline:run

PR #327, opened 2024-10-21
Refactor Federation Watcher

apancorb:
### Summary

The idea with this PR is to refactor the main Federation Watcher class an put it into its own package called `watcher`. This would be nice since the current state is a bit hard to understand how it works. Maybe this can help in the future and make future changes easier to implement.


marcos-iov (commented):
Left some suggestions/ideas. Let me know if you'd like to discuss further

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+        });
+    }
+
+    /**
+     * Updates the current state of the federations by checking if the active or
+     * retiring federations have changed. If a federation change is detected, it notifies
+     * the {@code FederationWatcherListener}.
+     */
+    public void updateState() {
```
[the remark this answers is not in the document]
apancorb: That is a good question. It should not be with the changes now. [Before](https://github.com/rsksmart/powpeg-node/pull/327/files#diff-eee97682409d79c95cbd6586ebe72e6ea347c78fca3099803a616225986b5ce6L366) it made sense since it was calling it manually, but I think this is redundant and should already be doing it (at least btcreleaseclient doesn't do it). If that is okay, we can make this private. wdyt?

src/main/java/co/rsk/federate/watcher/FederationWatcherListener.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Invoked when the retiring federation changes.
+     *
+     * @param oldFederation an {@code Optional} containing the previous retiring federation.
+     *                      This will be empty if there was no retiring federation before.
+     * @param newFederation an {@code Optional} containing the new retiring federation.
+     *                      This will be empty if there is no retiring federation after the change.
+     */
+    void onRetiringFederationChange(Optional<Federation> oldFederation, Optional<Federation> newFederation);
```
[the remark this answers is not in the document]
apancorb: I have redesigned the interface and removed the oldFederation param since it is only being used for a debug logging statement which is already being donde before anyways. This is more simple to reason. Let me know what you think.

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public void onRetiringFederationChange(Optional<Federation> oldFederation, Optional<Federation> newFederation) {
+        String oldFederationAddress = oldFederation.map(federation -> federation.getAddress().toString()).orElse("NONE");
+        String newFederationAddress = newFederation.map(federation -> federation.getAddress().toString()).orElse("NONE");
+        logger.debug("[onRetiringFederationChange] Retiring federation change: from {} to {}", oldFederationAddress, newFederationAddress);
+        triggerClientChange(btcToRskClientRetiring, newFederation);
+    }
+
+    private void triggerClientChange(BtcToRskClient client, Optional<Federation> federation) {
```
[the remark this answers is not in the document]
apancorb: Updated

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
```diff
@@ -0,0 +1,78 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.Objects;
+import java.util.Optional;
+
+public class FederationWatcherListenerImpl implements FederationWatcherListener {
+
+    private static final Logger logger = LoggerFactory.getLogger(FederationWatcherListenerImpl.class);
+    
+    private final FederationMember federationMember;
```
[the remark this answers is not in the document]
apancorb: I think this is needed here, not sure how we could start the clients beforehand.

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(FederationWatcher.class);
+
+    private final Ethereum rsk;
+
+    private FederationProvider federationProvider;
+    private FederationWatcherListener federationWatcherListener; 
+
+    private Optional<Federation> activeFederation = Optional.empty();
+    private Optional<Federation> retiringFederation = Optional.empty();
```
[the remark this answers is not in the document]
apancorb: Updated to not use optionals.

src/main/java/co/rsk/federate/watcher/FederationWatcherListener.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Invoked when the active federation changes.
+     *
+     * @param oldFederation an {@code Optional} containing the previous active federation.
+     *                      This will be empty if there was no active federation before.
+     * @param newFederation an {@code Optional} containing the new active federation.
+     *                      This will be empty if there is no active federation after the change.
+     */
+    void onActiveFederationChange(Optional<Federation> oldFederation, Optional<Federation> newFederation);
```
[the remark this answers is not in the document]
apancorb: This is true, I have changed the interface design since if we know there is a change then we also know the value to be guranteed. So it does not make sense to have optionals as params. Let me know what you think.

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
```diff
@@ -0,0 +1,124 @@
...
+     * the {@code FederationWatcherListener}.
+     */
+    private void updateState() {
+        updateActiveFederation();
+        updateRetiringFederation();
+    }
+
+    private void updateActiveFederation() {
+        // Check if active federation has changed
+        Optional<Address> currentlyActiveFederationAddress = Optional.ofNullable(federationProvider.getActiveFederationAddress());
```
julia-zack, line 87: why optional? shouldn't always be an active federation?

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
```diff
@@ -0,0 +1,124 @@
...
+        Optional<Address> oldActiveFederationAddress = Optional.ofNullable(activeFederation).map(Federation::getAddress);
+        boolean hasActiveFederationChanged = !currentlyActiveFederationAddress.equals(oldActiveFederationAddress);
+
+        if (hasActiveFederationChanged) {
+            // Active federation changed
```
julia-zack, line 92: nit: considering the boolean, i think this comment is no needed :)

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
```diff
@@ -0,0 +1,124 @@
...
+            logger.info("[updateActiveFederation] Active federation changed from {} to {}",
+                oldActiveFederationAddress.orElse(null),
+                currentlyActiveFederationAddress.orElse(null));
```
julia-zack, line 95: same as above

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    private void updateRetiringFederation() {
+        // Check if retiring federation has changed
+        Optional<Address> currentlyRetiringFederationAddress = federationProvider.getRetiringFederationAddress();
+        Optional<Address> oldRetiringFederationAddress = Optional.ofNullable(retiringFederation).map(Federation::getAddress);
+        boolean hasRetiringFederationChanged = !currentlyRetiringFederationAddress.equals(oldRetiringFederationAddress);
+
+        if (hasRetiringFederationChanged) {
+            // Retiring federation changed
```
julia-zack, line 112: same for this comment :)

src/main/java/co/rsk/federate/watcher/FederationWatcherListener.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public interface FederationWatcherListener {
+
+    /**
+     * Invoked when the active federation changes.
+     *
+     * @param newFederation the new active federation after the change.
+     *                      This will never be {@code null}; the active federation is always present.
+     */
+    void onActiveFederationChange(Federation newFederation);
```
julia-zack, line 17:
```suggestion
    void onActiveFederationChange(Federation newActiveFederation);
```

maybe redundant, but at first reading it caused me some confusion. Just a suggestion, no big deal

src/main/java/co/rsk/federate/watcher/FederationWatcherListener.java
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Invoked when the retiring federation changes.
+     *
+     * @param newFederation the new retiring federation after the change.
+     *                      This can be {@code null}; the retiring federation is not always present.
+     */
+    void onRetiringFederationChange(Federation newFederation);
```
julia-zack, line 25:
```suggestion
    void onRetiringFederationChange(Federation newRetiringFederation);
```

same here

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
```diff
@@ -0,0 +1,124 @@
...
+                // until this change effectively becomes a part of RSK's "reality".
+                // The case in which we go back and forth to and from a sidechain in which the
+                // federation effectively changed is still to be explored deeply, but the same reasoning
+                // should apply since going back and forth would trigger two federation changes.
+                // A client trying to send bitcoins to the new federation without waiting
+                // a good number of confirmations would be, essentially, "playing with fire".
+                logger.info("[updateState] New best block, updating state");
+                updateState();
+            }
+        });
```
julia-zack, line 72: not a big fan of having this block of code inside the `start` method. I think before was easier to follow

apancorb: The idea is to avoid nested inner clases if possible. This might seem big because of the comment. I have reorganized the comment to be more readable.

src/main/java/co/rsk/federate/watcher/FederationWatcherListener.java
apancorb: I think we can leave as is

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
apancorb: Okay removed the comments

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
apancorb: This is to do the equality check between the optionals.

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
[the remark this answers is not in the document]
nathanieliov: I see this check it's been done twice. Here in the`FederationWatcherListenerImpl` and also [here](https://github.com/rsksmart/powpeg-node/blob/c5887afe30b010f9dd0a1c24855051f504d18043/src/main/java/co/rsk/federate/BtcToRskClient.java#L131) in`BtcToRskClient`

Actually, it is using the same FedMember instance set in these places:  [FedNodeRunner.startFederate](https://github.com/rsksmart/powpeg-node/blob/76580d6b28563c68112981393ee49630f8239995/src/main/java/co/rsk/federate/FedNodeRunner.java#L188) and [FedNodeRunner.configureFederateSupport](https://github.com/rsksmart/powpeg-node/blob/76580d6b28563c68112981393ee49630f8239995/src/main/java/co/rsk/federate/FedNodeRunner.java#L348)

Having said that, I think we can do what Marcos says.

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
```diff
@@ -0,0 +1,124 @@
...
+
+            federationWatcherListener.onActiveFederationChange(currentlyActiveFederation);
+            this.activeFederation = currentlyActiveFederation;
+        }
+    }
+
+    private void updateRetiringFederation() {
+        Optional<Address> currentlyRetiringFederationAddress = federationProvider.getRetiringFederationAddress();
+        Optional<Address> oldRetiringFederationAddress = Optional.ofNullable(retiringFederation).map(Federation::getAddress);
+
```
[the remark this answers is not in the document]
apancorb: Here we need it, consider the case of having a current retiring federation that is null and an old one that is not null. You would have a NPE when doing the equality check here. This is why the old code does the equality check using the wrapped Optionals.

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
apancorb: The issue arises if for instance activeFederation is null. This line `Address oldActiveFederationAddress = activeFederation.getAddress();`,  could throw an NPE if activeFederation is null. Wrapping in an Optional avoids this case. Federation provider should always return a not null federation, but we still wrap it in an Optional just to make it consistent with the Retiring Federation implementation which is not always the case.

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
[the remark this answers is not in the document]
apancorb: I see, since this check is already being done [here](https://github.com/rsksmart/powpeg-node/blob/c5887afe30b010f9dd0a1c24855051f504d18043/src/main/java/co/rsk/federate/BtcToRskClient.java#L131), and we calling the start method from the triggerClientChange, we can avoid having this field and this check. Does that sound good?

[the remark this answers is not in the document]
apancorb: Acctualy just realized it we would still need to do that check since if we just let it to be done for `btcToRskClient` we would be starting the `btcReleaseClient` even if it is not part of the new federation.

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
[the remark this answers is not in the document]
nathanieliov: You are right! Thanks for the explanation. Very good catch.

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
```diff
@@ -0,0 +1,67 @@
...
+                "[triggerClientChange] Joined {} federation", newFederation.getAddress());
+        } catch (Exception e) {
+            logger.error(
+                "[triggerClientChange] This federation ({}) cannot be started: {}",
+                newFederation.getAddress(),
+                e.getMessage());
+        }
+    }
+    
+    private void triggerClearingRetiringFederationClient() {
```
apancorb, line 64: Open to better names :)

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
```diff
@@ -0,0 +1,67 @@
...
+        Objects.requireNonNull(newFederation);
+      
+        try {
+            // Stop the current clients
+            btcToRskClient.stop();
+            btcReleaseClient.stop(newFederation);
+
+            // Start the current clients
+            btcToRskClient.start(newFederation);
+            btcReleaseClient.start(newFederation);
```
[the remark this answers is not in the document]
nathanieliov: Didn't follow. Didn't understand the issue.

src/main/java/co/rsk/federate/btcreleaseclient/BtcReleaseClient.java
```diff
@@ -160,10 +161,21 @@ public void setup(
...
+
         if (!observedFederations.contains(federation)) {
             observedFederations.add(federation);
-            logger.debug("[start] observing Federation {}", federation.getAddress());
+            logger.debug("[start] Observing federation {}", federation.getAddress());
         }
+
         if (observedFederations.size() == 1) {
             // If there is just one observed Federation, it means the btcReleaseClient wasn't started
             logger.debug("[start] Starting");
```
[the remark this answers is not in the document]
apancorb: Since the listener just gets activated once there is one observed federation available. Consider the case you start it again with a different federation. Then you will have two listener instances running in the background doing the same computation twice. Thus, when size is 1 it means the listener for the RSK blockchain gets activated, if it decreases to zero then there is no need for the listener to be running.

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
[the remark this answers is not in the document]
apancorb: Notice how for btcReleaseClient we stop and start using the newFederation. For me it does not make sense, it should stop the old one and start the new federation.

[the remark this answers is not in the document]
nathanieliov: I see. Let's go deeper into this later. Create a task for it.

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
julia-zack: i still don't like it. If is null, we should throw an exception (or just return with a logged error) either before passing the value or in this method. I would avoid ending with a null active federation (not sure if this could happen but)

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
```diff
@@ -0,0 +1,67 @@
...
+        triggerClientChange(btcToRskClientActive, newActiveFederation);
+    }
+
+    @Override
+    public void onRetiringFederationChange(Federation newRetiringFederation) {
+        if (newRetiringFederation == null) {
+            triggerClearingRetiringFederationClient();
+        } else {
+            triggerClientChange(btcToRskClientRetiring, newRetiringFederation);
+        }
```
julia-zack, line 38:
```suggestion
        if (newRetiringFederation == null) {
            triggerClearingRetiringFederationClient();
            return;
        }
        triggerClientChange(btcToRskClientRetiring, newRetiringFederation);
```

wdyt?

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
```diff
@@ -0,0 +1,67 @@
...
+                "[triggerClientChange] Joined {} federation", newFederation.getAddress());
+        } catch (Exception e) {
+            logger.error(
+                "[triggerClientChange] This federation ({}) cannot be started: {}",
+                newFederation.getAddress(),
+                e.getMessage());
+        }
+    }
+    
+    private void triggerClearingRetiringFederationClient() {
```
julia-zack, line 64:
```suggestion
    private void clearRetiringFederationClient() {
```
wdyt? Just because you asked for suggestions jaja

apancorb: Sounds good :)

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
apancorb: Alright

src/main/java/co/rsk/federate/watcher/FederationWatcher.java
apancorb: You can have an active federation as null at node start time. I have changed it to the original implementation. Let me know if this is fine. Thanks!

julia-zack (approved):

src/main/java/co/rsk/federate/btcreleaseclient/BtcReleaseClient.java
```diff
@@ -160,10 +161,21 @@ public void setup(
     }
 
     public void start(Federation federation) {
+       FederationMember federationMember = federatorSupport.getFederationMember();
+       if (federation.isMember(federationMember)) {
```
marcos-iov, line 165: Should this condition be negative? `if (!isMember)` 😳

Isn't this covered by a test?

src/main/java/co/rsk/federate/watcher/FederationWatcherListenerImpl.java
```diff
@@ -0,0 +1,71 @@
+package co.rsk.federate.watcher;
+
+import co.rsk.federate.BtcToRskClient;
+import co.rsk.federate.btcreleaseclient.BtcReleaseClient;
+import co.rsk.peg.federation.Federation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.Objects;
+
+public class FederationWatcherListenerImpl implements FederationWatcherListener {
```
marcos-iov, line 10: Shouldn't this class have tests as well?

apancorb: Added unit tests, let me know if it looks good

marcos-iov (approved):

PR #329, opened 2024-10-25
Implement getProposedFederation in FederationProviderFromFederatorSupport

apancorb:
### Summary

This PR includes the implementation of `getProposedFederation`. As well as bridge RPC methods for specific proposed federation implementations for `getFederationSize`, `getFederatorPublicKeyOfType`,`getFederationCreationTime`, and `getFederationCreationBlockNumber`.

### Motivation and Context

https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP419.md

### DIFF
--- a/src/main/java/co/rsk/federate/FederationProvider.java
+++ b/src/main/java/co/rsk/federate/FederationProvider.java
@@ -29,7 +29,7 @@ import java.util.Optional;
 public interface FederationProvider {
 
     /**
-     * Retrieves the currently active federation, which is responsible for processing transactions.
+     * Retrieves the currently active federation.
      *
      * @return the active {@link Federation} instance
      */
@@ -58,7 +58,7 @@ public interface FederationProvider {
     Optional<Address> getRetiringFederationAddress();
 
     /**
-     * Retrieves the currently proposed federation, if one exists. This federation is awaiting activation.
+     * Retrieves the currently proposed federation, if one exists. This federation is awaiting validation.
      *
      * @return an {@link Optional} containing the proposed {@link Federation}, or {@link Optional#empty()} if none exists
      */
--- a/src/main/java/co/rsk/federate/FederationProvider.java
+++ b/src/main/java/co/rsk/federate/FederationProvider.java
@@ -22,24 +22,52 @@ import co.rsk.peg.federation.Federation;
 import java.util.Optional;
 
 /**
- * Implementors of this interface must be able to provide
- * federation instances.
- *
- * @author Ariel Mendelzon
+ * Provides access to various federation instances, including the active, retiring, and proposed federations.
+ * Implementations of this interface must define how to retrieve the federations and their respective addresses.
+ * These methods allow clients to manage and monitor federations in different lifecycle stages within the bridge.
  */
 public interface FederationProvider {
-    // The currently "active" federation
+
+    /**
+     * Retrieves the currently active federation, which is responsible for processing transactions.
+     *
+     * @return the active {@link Federation} instance
+     */
     Federation getActiveFederation();
-    // The currently "active" federation's address
+
+    /**
+     * Retrieves the address of the currently active federation.
+     *
+     * @return the {@link Address} of the active federation
+     */
     Address getActiveFederationAddress();
 
-    // The currently "retiring" federation
+    /**
+     * Retrieves the currently retiring federation, if one exists. This federation is in transition 
+     * and will soon be replaced by a new active federation.
+     *
+     * @return an {@link Optional} containing the retiring {@link Federation}, or {@link Optional#empty()} if none exists
+     */
     Optional<Federation> getRetiringFederation();
-    // The currently "retiring" federation's address
+
+    /**
+     * Retrieves the address of the currently retiring federation, if one exists.
+     *
+     * @return an {@link Optional} containing the {@link Address} of the retiring federation, or {@link Optional#empty()} if none exists
+     */
     Optional<Address> getRetiringFederationAddress();
 
-    // The currently "proposed" federation
+    /**
+     * Retrieves the currently proposed federation, if one exists. This federation is awaiting activation.
+     *
+     * @return an {@link Optional} containing the proposed {@link Federation}, or {@link Optional#empty()} if none exists
+     */
     Optional<Federation> getProposedFederation();
-    // The currently "proposed" federation's address
+
+    /**
+     * Retrieves the address of the currently proposed federation, if one exists.
+     *
+     * @return an {@link Optional} containing the {@link Address} of the proposed federation, or {@link Optional#empty()} if none exists
+     */
     Optional<Address> getProposedFederationAddress();
 }
--- a/src/main/java/co/rsk/federate/FederationProvider.java
+++ b/src/main/java/co/rsk/federate/FederationProvider.java
@@ -38,6 +38,8 @@ public interface FederationProvider {
     // The currently "retiring" federation's address
     Optional<Address> getRetiringFederationAddress();
 
+    // The currently "proposed" federation
+    Optional<Federation> getProposedFederation();
     // The currently "proposed" federation's address
     Optional<Address> getProposedFederationAddress();
 }
--- a/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
+++ b/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
@@ -157,20 +157,20 @@ public class FederationProviderFromFederatorSupport implements FederationProvide
                 federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.BTC)
                     .map(ECKey::getPubKey)
                     .map(BtcECKey::fromPublicOnly)
-                    .orElseThrow(() -> new IllegalStateException()),
+                    .orElseThrow(IllegalStateException::new),
                 federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.RSK)
-                    .orElseThrow(() -> new IllegalStateException()),
+                    .orElseThrow(IllegalStateException::new),
                 federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.MST)
-                    .orElseThrow(() -> new IllegalStateException())
+                    .orElseThrow(IllegalStateException::new)
             ))
             .toList();
 
         FederationArgs federationArgs = new FederationArgs(
             federationMembers,
             federatorSupport.getProposedFederationCreationTime()
-                .orElseThrow(() -> new IllegalStateException()),
+                .orElseThrow(IllegalStateException::new),
             federatorSupport.getProposedFederationCreationBlockNumber()
-                .orElseThrow(() -> new IllegalStateException()),
+                .orElseThrow(IllegalStateException::new),
             federatorSupport.getBtcParams()
         );
 
--- a/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
+++ b/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
@@ -152,7 +152,7 @@ public class FederationProviderFromFederatorSupport implements FederationProvide
             return Optional.empty();
         }
 
-        List<FederationMember> members = IntStream.range(0, federationSize)
+        List<FederationMember> federationMembers = IntStream.range(0, federationSize)
             .mapToObj(i -> new FederationMember(
                 federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.BTC)
                     .map(ECKey::getPubKey)
@@ -166,7 +166,7 @@ public class FederationProviderFromFederatorSupport implements FederationProvide
             .toList();
 
         FederationArgs federationArgs = new FederationArgs(
-            members,
+            federationMembers,
             federatorSupport.getProposedFederationCreationTime()
                 .orElseThrow(() -> new IllegalStateException()),
             federatorSupport.getProposedFederationCreationBlockNumber()
@@ -174,10 +174,12 @@ public class FederationProviderFromFederatorSupport implements FederationProvide
             federatorSupport.getBtcParams()
         );
 
-        Federation initialFederation = FederationFactory.buildStandardMultiSigFederation(federationArgs);
-        Federation expectedFederation = getExpectedFederation(initialFederation, proposedFederationAddress.get());
+        Federation federation = FederationFactory.buildP2shErpFederation(
+            federationArgs,
+            federationConstants.getErpFedPubKeysList(),
+            federationConstants.getErpFedActivationDelay());
 
-        return Optional.of(expectedFederation);
+        return Optional.of(federation);
     }
 
     @Override
--- a/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
+++ b/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
@@ -184,8 +184,9 @@ public class FederationProviderFromFederatorSupport implements FederationProvide
                 federationConstants.getErpFedActivationDelay());
 
             return Optional.of(federation);
-        } catch (IllegalStateException e) {
+        } catch (Exception e) {
             logger.error("[getProposedFederation] Unable to build the proposed federation", e);
+
             return Optional.empty();
         }
     }
--- a/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
+++ b/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
@@ -185,7 +185,7 @@ public class FederationProviderFromFederatorSupport implements FederationProvide
     @Override
     public Optional<Address> getProposedFederationAddress() {
         return Optional.of(federatorSupport)
-            .filter(federatorSupport -> federatorSupport.getConfigForBestBlock().isActive(RSKIP419))
+            .filter(fedSupport -> fedSupport.getConfigForBestBlock().isActive(RSKIP419))
             .flatMap(FederatorSupport::getProposedFederationAddress);
     }
 
--- a/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
+++ b/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
@@ -17,7 +17,10 @@
  */
 package co.rsk.federate;
 
+import static co.rsk.peg.federation.FederationChangeResponseCode.FEDERATION_NON_EXISTENT;
+import static co.rsk.peg.federation.FederationMember.KeyType;
 import static org.ethereum.config.blockchain.upgrades.ConsensusRule.RSKIP123;
+import static org.ethereum.config.blockchain.upgrades.ConsensusRule.RSKIP417;
 
 import co.rsk.bitcoinj.core.Address;
 import co.rsk.bitcoinj.core.BtcECKey;
@@ -30,6 +33,7 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.IntStream;
 
 /**
  * Provides a federation using a FederatorSupport instance, which in turn
@@ -136,8 +140,53 @@ public class FederationProviderFromFederatorSupport implements FederationProvide
     }
 
     @Override
+    public Optional<Federation> getProposedFederation() {
+        if (!federatorSupport.getConfigForBestBlock().isActive(RSKIP417)) {
+            return Optional.empty();
+        }
+       
+        Optional<Address> proposedFederationAddress = getProposedFederationAddress();
+        if (proposedFederationAddress.isEmpty()) {
+            return Optional.empty();
+        }
+
+        Integer federationSize = federatorSupport.getProposedFederationSize()
+            .orElse(FEDERATION_NON_EXISTENT.getCode());
+        if (federationSize == FEDERATION_NON_EXISTENT.getCode()) {
+            return Optional.empty();
+        }
+
+        List<FederationMember> members = IntStream.range(0, federationSize)
+            .mapToObj(i -> new FederationMember(
+                federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.BTC)
+                    .map(ECKey::getPubKey)
+                    .map(BtcECKey::fromPublicOnly)
+                    .orElse(null),
+                federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.RSK)
+                    .orElse(null),
+                federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.MST)
+                    .orElse(null)
+            ))
+            .toList();
+
+        FederationArgs federationArgs = new FederationArgs(
+            members,
+            federatorSupport.getProposedFederationCreationTime().orElse(null),
+            federatorSupport.getProposedFederationCreationBlockNumber().orElse(null),
+            federatorSupport.getBtcParams()
+        );
+
+        Federation initialFederation = FederationFactory.buildStandardMultiSigFederation(federationArgs);
+        Federation expectedFederation = getExpectedFederation(initialFederation, proposedFederationAddress.get());
+
+        return Optional.of(expectedFederation);
+    }
+
+    @Override
     public Optional<Address> getProposedFederationAddress() {
-        return federatorSupport.getProposedFederationAddress();
+        return Optional.of(federatorSupport)
+            .filter(federatorSupport -> federatorSupport.getConfigForBestBlock().isActive(RSKIP417))
+            .flatMap(FederatorSupport::getProposedFederationAddress);
     }
 
     private Federation getExpectedFederation(Federation initialFederation, Address expectedFederationAddress) {
--- a/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
+++ b/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
@@ -145,10 +145,9 @@ public class FederationProviderFromFederatorSupport implements FederationProvide
             return Optional.empty();
         }
        
-        Optional<Address> proposedFederationAddress = getProposedFederationAddress();
         Integer federationSize = federatorSupport.getProposedFederationSize()
             .orElse(FEDERATION_NON_EXISTENT.getCode());
-        if (federationSize == FEDERATION_NON_EXISTENT.getCode() || proposedFederationAddress.isEmpty()) {
+        if (federationSize == FEDERATION_NON_EXISTENT.getCode()) {
             return Optional.empty();
         }
 
--- a/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
+++ b/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
@@ -157,18 +157,20 @@ public class FederationProviderFromFederatorSupport implements FederationProvide
                 federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.BTC)
                     .map(ECKey::getPubKey)
                     .map(BtcECKey::fromPublicOnly)
-                    .orElse(null),
+                    .orElseThrow(() -> new IllegalStateException()),
                 federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.RSK)
-                    .orElse(null),
+                    .orElseThrow(() -> new IllegalStateException()),
                 federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.MST)
-                    .orElse(null)
+                    .orElseThrow(() -> new IllegalStateException())
             ))
             .toList();
 
         FederationArgs federationArgs = new FederationArgs(
             members,
-            federatorSupport.getProposedFederationCreationTime().orElse(null),
-            federatorSupport.getProposedFederationCreationBlockNumber().orElse(null),
+            federatorSupport.getProposedFederationCreationTime()
+                .orElseThrow(() -> new IllegalStateException()),
+            federatorSupport.getProposedFederationCreationBlockNumber()
+                .orElseThrow(() -> new IllegalStateException()),
             federatorSupport.getBtcParams()
         );
 
--- a/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
+++ b/src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
@@ -37,11 +37,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Provides a federation using a FederatorSupport instance, which in turn
- * gathers the federation from the bridge contract of the ethereum
- * network it is attached to.
+ * A provider that supplies the current active, retiring, and proposed federations by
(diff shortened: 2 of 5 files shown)


src/main/java/co/rsk/federate/FederationProvider.java
```diff
@@ -39,6 +39,8 @@ public interface FederationProvider {
     // The currently "retiring" federation's address
     Optional<Address> getRetiringFederationAddress();
 
+    // The currently "proposed" federation
```
jeremy-then, line 42: Do we really need all these simple comments on top of the method definition? There´s no logic here. I think they are not adding much value really.

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -93,7 +97,7 @@ public Optional<Federation> getRetiringFederation() {
         Integer federationSize = federatorSupport.getRetiringFederationSize();
         Optional<Address> optionalRetiringFederationAddress = getRetiringFederationAddress();
 
-        if (federationSize == -1 || !optionalRetiringFederationAddress.isPresent()) {
+        if (federationSize == FEDERATION_NON_EXISTENT.getCode() || !optionalRetiringFederationAddress.isPresent()) {
```
nathanieliov, line 100:
```suggestion
        if (federationSize == FEDERATION_NON_EXISTENT.getCode() || optionalRetiringFederationAddress.isEmpty()) {
```

Please check IDE warnings and suggestions

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -135,9 +139,50 @@ public Optional<Address> getRetiringFederationAddress() {
...
+        if (federationSize == FEDERATION_NON_EXISTENT.getCode() || proposedFederationAddress.isEmpty()) {
+            return Optional.empty();
+        }
+
+        List<FederationMember> members = IntStream.range(0, federationSize)
+            .mapToObj(i -> new FederationMember(
+                federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.BTC)
+                    .map(ECKey::getPubKey)
+                    .map(BtcECKey::fromPublicOnly)
+                    .orElse(null),
```
nathanieliov, line 160: It makes no sense to create fed member with null values. It should throw an `IllegalStateException` instead, as once you check a proposed federation exists, you expect its keys to be obtained.

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -135,9 +139,50 @@ public Optional<Address> getRetiringFederationAddress() {
...
+                federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.RSK)
```
nathanieliov, line 161: Same

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -135,9 +139,50 @@ public Optional<Address> getRetiringFederationAddress() {
...
+                    .orElse(null),
+                federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.MST)
+                    .orElse(null)
```
nathanieliov, line 164: And same

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -135,9 +139,50 @@ public Optional<Address> getRetiringFederationAddress() {
...
     }
 
+    @Override
+    public Optional<Federation> getProposedFederation() {
+        if (!federatorSupport.getConfigForBestBlock().isActive(RSKIP419)) {
+            return Optional.empty();
+        }
+       
+        Optional<Address> proposedFederationAddress = getProposedFederationAddress();
+        Integer federationSize = federatorSupport.getProposedFederationSize()
```
nathanieliov, line 149: I think you don't need the size to check if a proposedFederation exists.

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -135,9 +139,50 @@ public Optional<Address> getRetiringFederationAddress() {
...
+                federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.RSK)
+                    .orElse(null),
+                federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.MST)
+                    .orElse(null)
+            ))
+            .toList();
+
+        FederationArgs federationArgs = new FederationArgs(
+            members,
+            federatorSupport.getProposedFederationCreationTime().orElse(null),
```
nathanieliov, line 170: Same, I would avoid this logic passing null.

src/main/java/co/rsk/federate/FederatorSupport.java
nathanieliov, line 1: Check IDE warnings in this file

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -135,9 +139,50 @@ public Optional<Address> getRetiringFederationAddress() {
...
+            federatorSupport.getProposedFederationCreationBlockNumber().orElse(null),
+            federatorSupport.getBtcParams()
+        );
+
+        Federation initialFederation = FederationFactory.buildStandardMultiSigFederation(federationArgs);
+        Federation expectedFederation = getExpectedFederation(initialFederation, proposedFederationAddress.get());
```
nathanieliov, line 176: I see how `getExpectedFederation` builds the federation, and I think this method is unsuitable for this case. At the beginning, you have already checked if `RSKIP419` is active, so we can be sure that the proposed federation will be a `P2shFederation`. if not, I would like to understand why you would be a different type?

src/test/java/co/rsk/federate/FederationProviderFromFederatorSupportTest.java
```diff
@@ -370,9 +371,159 @@ void getRetiringFederation_present_p2sh_erp_federation() {
...
+            .thenReturn(Optional.of(9));
+
+        // Act & Assert
+        assertEquals(Optional.empty(), federationProvider.getProposedFederation());
+        verify(federatorSupportMock).getProposedFederationSize();
+        verify(federatorSupportMock).getProposedFederationAddress();
+    }
+
+    @Test
+    void getProposedFederation_whenExistsAndIsStandardMultisigFederation_shouldReturnProposedFederation() {
```
nathanieliov, line 416: As said above, once `RSKIP419`, all pending and proposed federations are always `P2shFeds`. Therefore, this is an unrealistic scenario.

src/test/java/co/rsk/federate/FederationProviderFromFederatorSupportTest.java
```diff
@@ -370,9 +371,159 @@ void getRetiringFederation_present_p2sh_erp_federation() {
...
+
+        // Assert
+        assertTrue(proposedFederation.isPresent());
+        assertEquals(STANDARD_MULTISIG_FEDERATION_FORMAT_VERSION, proposedFederation.get().getFormatVersion());
+        assertEquals(expectedFederation, proposedFederation.get());
+        assertEquals(expectedFederationAddress, proposedFederation.get().getAddress());
+    }
+
+    @Test
+    void getProposedFederation_whenExistsAndIsNonStandardErpFederation_shouldReturnProposedFederation() {
```
nathanieliov, line 451: Same

src/main/java/co/rsk/federate/FederationProvider.java
apancorb: I agree, I have made better javadocs for the interface. Let me know what you think

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
apancorb: Thanks, updated

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
apancorb: I agree, but since this is also being done in the retiring federation impl I left it as is. Do you recommend removing it?

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
apancorb: I agree, I have changed it to throw an IllegalStateException if this happens. Let me know what you think.

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
apancorb: Thanks for pointing this out :) , I have made the changes. Let me know if this looks good now.

src/test/java/co/rsk/federate/FederationProviderFromFederatorSupportTest.java
apancorb: Removed, as well as the other one

src/main/java/co/rsk/federate/FederatorSupport.java
apancorb: What warning are you referring to?

src/main/java/co/rsk/federate/FederationProvider.java
```diff
@@ -23,27 +23,60 @@
...
- * @author Ariel Mendelzon
+ * Provides access to various federation instances, including the active, retiring, and proposed federations.
+ * Implementations of this interface must define how to retrieve the federations and their respective addresses.
+ * These methods allow clients to manage and monitor federations in different lifecycle stages within the bridge.
  */
 public interface FederationProvider {
-    // The currently "active" federation
+
+    /**
+     * Retrieves the currently active federation, which is responsible for processing transactions.
```
julia-zack, line 33:
```suggestion
     * Retrieves the currently active federation.
```

i think theres no need to clarify anything here

julia-zack: also i find `which is responsible for processing transactions` kind of vague, which transactions? wdym by processing? etc

src/main/java/co/rsk/federate/FederationProvider.java
```diff
@@ -23,27 +23,60 @@
...
+    /**
+     * Retrieves the address of the currently retiring federation, if one exists.
+     *
+     * @return an {@link Optional} containing the {@link Address} of the retiring federation, or {@link Optional#empty()} if none exists
+     */
     Optional<Address> getRetiringFederationAddress();
 
-    // The currently "proposed" federation's address
+    /**
+     * Retrieves the currently proposed federation, if one exists. This federation is awaiting activation.
```
julia-zack, line 62:
```suggestion
     * Retrieves the currently proposed federation, if one exists. This federation is awaiting validation.
```

src/main/java/co/rsk/federate/FederationProvider.java
```diff
@@ -23,27 +23,60 @@
...
-    // The federations that are "live", that is, are still
-    // operational. This should be the active federation
-    // plus the retiring federation, if one exists
+    /**
+     * Retrieves a list of live federations, which includes both the active federation 
+     * and the retiring federation, if one exists.
+     *
+     * @return a {@link List} of live {@link Federation} instances
+     */
     List<Federation> getLiveFederations();
```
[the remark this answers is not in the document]
julia-zack: Lets create a task for that

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
julia-zack: actually, since you are using the fed size below, i think we could just check for it and remove the address check

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -135,9 +139,54 @@ public Optional<Address> getRetiringFederationAddress() {
...
 
+    @Override
+    public Optional<Federation> getProposedFederation() {
+        if (!federatorSupport.getConfigForBestBlock().isActive(RSKIP419)) {
+            return Optional.empty();
+        }
+       
+        Optional<Address> proposedFederationAddress = getProposedFederationAddress();
+        Integer federationSize = federatorSupport.getProposedFederationSize()
+            .orElse(FEDERATION_NON_EXISTENT.getCode());
```
julia-zack, line 150: i think this should be being done in the bridge, [as we are doing with the retiring fed](https://github.com/rsksmart/rskj/blob/feature/powpeg_validation_protocol-phase4/rskj-core/src/main/java/co/rsk/peg/federation/FederationSupportImpl.java#L237)

wdyt? The fed support should just return what it gets from the bridge.

src/main/java/co/rsk/federate/FederationProvider.java
apancorb: Sounds good, I have updated it.

src/main/java/co/rsk/federate/FederationProvider.java
apancorb: Nice, updated

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
apancorb: This is also being done in the Bridge for the proposed fed. The fed support also only returns what it gets from the Bridge, here we are explicitly defining how to treat an empty optional in case we receive it. wdyt?

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
apancorb: Alright I will delete the check for the address part then

src/main/java/co/rsk/federate/FederationProvider.java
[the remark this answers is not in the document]
nathanieliov: Agree

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -135,9 +139,53 @@ public Optional<Address> getRetiringFederationAddress() {
...
+                    .orElseThrow(IllegalStateException::new),
+                federatorSupport.getProposedFederatorPublicKeyOfType(i, KeyType.MST)
+                    .orElseThrow(IllegalStateException::new)
+            ))
+            .toList();
+
+        FederationArgs federationArgs = new FederationArgs(
+            federationMembers,
+            federatorSupport.getProposedFederationCreationTime()
+                .orElseThrow(IllegalStateException::new),
```
nathanieliov, line 170: Where is this exception being caught (if so)? I think we should add a context message when instantiating it or when catching it. Wdyt?

src/test/java/co/rsk/federate/FederationProviderFromFederatorSupportTest.java
```diff
@@ -370,9 +371,72 @@ void getRetiringFederation_present_p2sh_erp_federation() {
...
+            getFederationMembersFromPks(1, 1000, 2000, 3000, 4000, 5000));
+        Address expectedFederationAddress = expectedFederation.getAddress();
+        Integer federationSize = 5;
+        when(federatorSupportMock.getProposedFederationSize()).thenReturn(Optional.of(federationSize));
+        when(federatorSupportMock.getProposedFederationCreationTime()).thenReturn(Optional.of(creationTime));
+        when(federatorSupportMock.getProposedFederationAddress()).thenReturn(Optional.of(expectedFederationAddress));
+        when(federatorSupportMock.getBtcParams()).thenReturn(testnetParams);
+        when(federatorSupportMock.getProposedFederationCreationBlockNumber()).thenReturn(Optional.of(0L));
+        for (int i = 0; i < federationSize; i++) {
+            when(federatorSupportMock.getProposedFederatorPublicKeyOfType(i, FederationMember.KeyType.BTC))
```
nathanieliov, line 416: Test case when the key does not exist and it throws an `IllegalStateException`, wdyt?

src/test/java/co/rsk/federate/FederatorSupportTest.java
```diff
@@ -44,6 +48,7 @@ class FederatorSupportTest {
     private static final NetworkParameters NETWORK_PARAMETERS = BridgeMainNetConstants.getInstance().getBtcParams();
     private static final List<BtcECKey> KEYS = BitcoinTestUtils.getBtcEcKeysFromSeeds(new String[]{"k1", "k2", "k3"}, true);
     private static final Address DEFAULT_ADDRESS = BitcoinTestUtils.createP2SHMultisigAddress(NETWORK_PARAMETERS, KEYS);
+    private static final byte[] PUBLIC_KEY = Hex.decode("0497466f2b32bc3bb76d4741ae51cd1d8578b48d3f1e68da206d47321aec267ce78549b514e4453d74ef11b0cd5e4e4c364effddac8b51bcfc8de80682f952896f");
```
nathanieliov, line 51: Could not this be created from a BigInteger?

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
apancorb: True, I have wrapped in a try-catch with logging the error.

src/test/java/co/rsk/federate/FederationProviderFromFederatorSupportTest.java
apancorb: Updated

src/test/java/co/rsk/federate/FederatorSupportTest.java
apancorb: No, it seems to not be able to decode it

apancorb: No, it seems to not be able to decode it

nathanieliov:
```suggestion
private static final byte[] PUBLIC_KEY = ECKey.fromPrivate(BigInteger.valueOf(100)).getPubKey();
```

Try this

nathanieliov (approved):

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -135,9 +143,58 @@ public Optional<Address> getRetiringFederationAddress() {
         return federatorSupport.getRetiringFederationAddress();
     }
 
+    @Override
+    public Optional<Federation> getProposedFederation() {
+        if (!federatorSupport.getConfigForBestBlock().isActive(RSKIP419)) {
+            return Optional.empty();
+        }
```
julia-zack, line 150: this is not necessary. If RSKIP419 is not activated, the Bridge won't return anything

src/test/java/co/rsk/federate/FederatorSupportTest.java
```diff
@@ -186,6 +191,139 @@ void getProposedFederation_whenAddressStringIsValid_shouldReturnAddress() {
...
+        // Act
+        Optional<Integer> result = federatorSupport.getProposedFederationSize();
+
+        // Assert
+        assertTrue(result.isPresent());
+        assertEquals(expectedSize.intValue(), result.get());
+    }
+
+    @Test
+    void getProposedFederatorPublicKeyOfType_whenPublicKeyBytesIsNull_shouldReturnEmptyOptional() {
```
julia-zack, line 223:
```suggestion
    void getProposedFederatorPublicKeyOfType_whenPublicKeyIsNull_shouldReturnEmptyOptional() {
```

src/test/java/co/rsk/federate/FederatorSupportTest.java
```diff
@@ -186,6 +191,139 @@ void getProposedFederation_whenAddressStringIsValid_shouldReturnAddress() {
...
+
+        // Act
+        Optional<ECKey> result = federatorSupport.getProposedFederatorPublicKeyOfType(index, keyType);
+
+        // Assert
+        assertTrue(result.isEmpty());
+    }
+
+    @Test
+    void getProposedFederatorPublicKeyOfType_whenPublicKeyBytesArePresent_shouldReturnECKey() {
```
julia-zack, line 241:
```suggestion
    void getProposedFederatorPublicKeyOfType_whenPublicKeyArePresent_shouldReturnECKey() {
```

src/test/java/co/rsk/federate/FederatorSupportTest.java
```diff
@@ -186,6 +191,139 @@ void getProposedFederation_whenAddressStringIsValid_shouldReturnAddress() {
...
+        assertArrayEquals(expectedKey.getPubKey(), result.get().getPubKey());
+    }
+
+    @Test
+    void getProposedFederatorPublicKeyOfType_whenKeyTypeIsNull_shouldThrowNullPointerException() {
+        // Arrange
+        int index = 0;
+        
+        // Act & Assert
+        assertThrows(NullPointerException.class, () -> {
```
julia-zack, line 266: is this what we expect? a npe?

julia-zack: i mean, is it the best thing to do?

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
apancorb: If we include it we can avoid doing a callTx to the Bridge. Maybe is a better idea to leave it in place. wdyt?

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -135,9 +143,58 @@ public Optional<Address> getRetiringFederationAddress() {
...
+       
+        Integer federationSize = federatorSupport.getProposedFederationSize()
+            .orElse(FEDERATION_NON_EXISTENT.getCode());
+        if (federationSize == FEDERATION_NON_EXISTENT.getCode()) {
+            return Optional.empty();
+        }
```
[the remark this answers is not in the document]
apancorb: This would work if the Bridge returned only a null value for the federation size if the proposed fed did not exist. But it can return FEDERATION_NON_EXISTENT which would be present in the optional. We would have to check this before continuing.

src/test/java/co/rsk/federate/FederatorSupportTest.java
apancorb: Updated

src/test/java/co/rsk/federate/FederatorSupportTest.java
apancorb: Yes, since we are requiring it to be non null it will throw an NPE. Check: https://docs.oracle.com/javase/10/docs/api/java/util/Objects.html#requireNonNull(T)

src/test/java/co/rsk/federate/FederatorSupportTest.java
apancorb: Updated

src/test/java/co/rsk/federate/FederatorSupportTest.java
apancorb: Updated

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
julia-zack: I think the less the powpeg knows about rskips, the better

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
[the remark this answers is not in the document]
julia-zack: pretty sure the `FEDERATION_NON_EXISTENT` code is not being used for the proposed fed, but for the pending and retiring.

src/test/java/co/rsk/federate/FederatorSupportTest.java
julia-zack: you right, thanks!

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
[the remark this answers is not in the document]
apancorb: It was implemented [here](https://github.com/rsksmart/rskj/pull/2794/files#diff-34109dfc71ff55ae26e9ede5ada468333341d0be091e537ad21ccd847c57351fR1082).

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
apancorb: Interesting, maybe we can discuss this with the team

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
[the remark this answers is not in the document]
julia-zack: my bad, you right. But in that case it wont be necessary to have an optional right? Because you are always getting a value from the Bridge

src/test/java/co/rsk/federate/FederationProviderFromFederatorSupportTest.java
```diff
@@ -370,9 +370,93 @@ void getRetiringFederation_present_p2sh_erp_federation() {
...
 
+    @Test
+    void getProposedFederation_whenRSKIP419IsNotActivated_shouldReturnEmptyOptional() {
+        // Arrange
+        ActivationConfig.ForBlock configMock = mock(ActivationConfig.ForBlock.class);
+        when(federatorSupportMock.getConfigForBestBlock()).thenReturn(configMock);
+
+        // Act & Assert
+        assertEquals(Optional.empty(), federationProvider.getProposedFederation());
+    }
```
julia-zack, line 381: probably this test can be removed

apancorb: That is right, thanks

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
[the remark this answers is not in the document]
apancorb: The Bridge is always returning something, but the result of evaluating the callTx can be null for variety of reasons. With the optional you avoid this error handling and then when it is time to consume it you explicitly declare how you want to interpret a missing value. In this case we want to reinforce that it should be FEDERATION_NON_EXISTENT code.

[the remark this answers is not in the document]
julia-zack: I get it. But i don't like that we are casting a null response (that will come from an error, since we are returning `-1` when the proposed fed does not exist) to `FEDERATION_NON_EXISTENT`. I find it confusing, makes me think that the Bridge will return an empty response when there is no proposed fed.

julia-zack (approved):

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -93,7 +100,7 @@ public Optional<Federation> getRetiringFederation() {
         Integer federationSize = federatorSupport.getRetiringFederationSize();
         Optional<Address> optionalRetiringFederationAddress = getRetiringFederationAddress();
 
-        if (federationSize == -1 || !optionalRetiringFederationAddress.isPresent()) {
+        if (federationSize == FEDERATION_NON_EXISTENT.getCode() || optionalRetiringFederationAddress.isEmpty()) {
```
marcos-iov, line 103: I know this already existed here, but probably only one of the checks is enough right? I'd just stick to checking the federation size

src/main/java/co/rsk/federate/FederatorSupport.java
```diff
@@ -261,6 +257,43 @@ public Optional<Address> getProposedFederationAddress() {
...
+            .map(ECKey::fromPublicOnly);
+    }
+
+    public Optional<Instant> getProposedFederationCreationTime() {
+        BigInteger creationTime = bridgeTransactionSender.callTx(
+            federatorAddress, Bridge.GET_PROPOSED_FEDERATION_CREATION_TIME);
+
+        return Optional.ofNullable(creationTime)
+            .map(BigInteger::longValue)
+            .map(Instant::ofEpochMilli);
```
marcos-iov, line 286: Let's agree on this @julia-zack and @jeremy-then. Mili or seconds?

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
apancorb: Sounds good I have removed the extra check. I placed the call after the check. I also modifed the corresponding tests. I deleted one since it didn't make sense to have it anymore IMO. Let me know if that is fine.

src/main/java/co/rsk/federate/FederatorSupport.java
julia-zack: I think we should go with seconds yes

src/main/java/co/rsk/federate/FederationProviderFromFederatorSupport.java
```diff
@@ -91,13 +98,12 @@ public Address getActiveFederationAddress() {
...
         Integer federationSize = federatorSupport.getRetiringFederationSize();
-        Optional<Address> optionalRetiringFederationAddress = getRetiringFederationAddress();
 
-        if (federationSize == -1 || !optionalRetiringFederationAddress.isPresent()) {
+        if (federationSize == FEDERATION_NON_EXISTENT.getCode()) {
             return Optional.empty();
         }
 
-        Address retiringFederationAddress = optionalRetiringFederationAddress.get();
+        Address retiringFederationAddress = getRetiringFederationAddress().orElse(null);
```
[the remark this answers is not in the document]
apancorb: That is true, I have updated it to throw an IllegalStateException and added a unit test for it. Let me know if this looks good.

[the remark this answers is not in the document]
marcos-iov: Looking good but not sure about catching the exception and returning empty. It is an unexpected state, should we maybe let the exception propagate so we are informed that something is wrong?

If the retiring federation exists (members > -1) and we can't get the address something is very wrong. Returning empty implies there is no retiring federation and I don't think that's the right behavior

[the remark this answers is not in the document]
apancorb: Yes we can do this and I checked that this should be fine to let it propagate since the listener is wrapped in a try-catch so it won't bring the node down:

```java
    private void scheduleListenerCallbacks(Consumer<EthereumListener> callback) {
             for (EthereumListener listener : listeners) {
                 try {
                    callback.accept(listener);
                } catch (Throwable e) {
                     logger.error("Listener callback failed with exception", e);
                    panicProcessor.panic("thread", String.format("Listener callback failed with exception %s", e.getMessage()));
                }
            }
        }
```
Just notice that it will keep failling per onBestBlock which might be a good thing.

I have also removed it then for the Proposed federation. And updated the tests involved with this.

marcos-iov (approved):

PR #344, opened 2024-11-07
Add description for isPegoutEnabled parameter

apancorb:
### Summary

As requested, we will introduce a comment explaining the purpose of the `federator.pegout.enabled` parameter.

### DIFF
--- a/src/main/java/co/rsk/federate/config/PowpegNodeConfigParameter.java
+++ b/src/main/java/co/rsk/federate/config/PowpegNodeConfigParameter.java
@@ -4,13 +4,15 @@ import java.util.Objects;
 import java.util.function.Function;
 
 public enum PowpegNodeConfigParameter {
-  FEDERATOR_ENABLED("federator.enabled", Boolean.toString(true)),
-  PEGOUT_ENABLED("federator.pegout.enabled", Boolean.toString(true)),
-  UPDATE_BRIDGE_TIMER_ENABLED("federator.updateBridgeTimerEnabled", Boolean.toString(true)),
-  UPDATE_BRIDGE_BTC_BLOCKCHAIN("federator.updateBridgeBtcBlockchain", Boolean.toString(true)),
-  UPDATE_BRIDGE_BTC_COINBASE_TRANSACTIONS("federator.updateBridgeBtcCoinbaseTransactions", Boolean.toString(true)),
-  UPDATE_BRIDGE_BTC_TRANSACTIONS("federator.updateBridgeBtcTransactions", Boolean.toString(true)),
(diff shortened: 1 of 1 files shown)


src/main/java/co/rsk/federate/config/PowpegNodeConfigParameter.java
```diff
@@ -5,6 +5,9 @@
 
 public enum PowpegNodeConfigParameter {
   FEDERATOR_ENABLED("federator.enabled", Boolean.toString(true)),
+  // when enabled the federator will be able to attempt signning
```
julia-zack, line 8:
```suggestion
  // when enabled the federator will be able to attempt signing
```

src/main/java/co/rsk/federate/config/PowpegNodeConfigParameter.java
```diff
@@ -5,6 +5,9 @@
...
+  // rsk transactions waiting for signatures reported by the Bridge,
+  // such as pegouts
```
julia-zack, line 10: not sure if this line adds value

apancorb: I think it does, maybe in the future we can document what each param does

src/main/java/co/rsk/federate/config/PowpegNodeConfigParameter.java
```diff
@@ -5,6 +5,9 @@
 
 public enum PowpegNodeConfigParameter {
   FEDERATOR_ENABLED("federator.enabled", Boolean.toString(true)),
```
jeremy-then, line 7: nit:
```suggestion
  FEDERATOR_ENABLED("federator.enabled", Boolean.TRUE.toString()),
```

jeremy-then (approved):
LGTM.

src/main/java/co/rsk/federate/config/PowpegNodeConfigParameter.java
julia-zack: i mean clarifying "such as pegouts", because atm is also enabling migrations for example. If we sign something else in the future we should remember to update the comment. Wdyt?

src/main/java/co/rsk/federate/config/PowpegNodeConfigParameter.java
apancorb: Nice, updated

src/main/java/co/rsk/federate/config/PowpegNodeConfigParameter.java
apancorb: I see, updated

julia-zack (approved):

marcos-iov (approved):

PR #396, opened 2025-02-17
Downgrade docker runtime to java 17 instead of 21

julia-zack:


src/main/resources/version.properties
```diff
@@ -1,2 +1,2 @@
-versionNumber='7.1.0.0'
```
marcos-iov, line 1: No need to upgrade the version, these changes are part of 7.1.0.0

marcos-iov (approved):

PR #412, opened 2025-03-17
Fix block selection. Remove HSM version 3

julia-zack:
**Background**
- The implementation of getConfirmedBlocks method was wrong: it was calculating the block difficulty considering uncles, and then applying the cap, instead of applying the cap in each uncle.
This was causing that the node thought it should send many more blocks to the HSM to reach the expected difficulty, ended up sending difficulty to advance aprox 1000 blocks, and our intention is for it to advance 100 blocks each time.

- On the other side, the HSM version 3 never really existed, so it didn’t have sense to have it as a possible HSM version. 

**Requirements**
- Fix block selection algorithm to apply the cap in each uncle and not at the end.
- Remove HSM version 3 as a possible version


marcos-iov (approved):

jeremy-then (approved):
LGTM.

PR #430, opened 2025-05-16
Remove signatures from transaction p2wsh p2sh

julianlen:


src/main/java/co/rsk/federate/btcreleaseclient/BtcReleaseClient.java
```diff
@@ -557,8 +557,9 @@ protected void removeSignaturesFromTransaction(BtcTransaction pegoutBtcTx, Feder
             Script inputRedeemScript = getRedeemScriptFromInput(txInput);
             logger.trace("[removeSignaturesFromTransaction] input {} scriptSig {}", inputIndex, pegoutBtcTx.getInput(inputIndex).getScriptSig());
             logger.trace("[removeSignaturesFromTransaction] input {} redeem script {}", inputIndex, inputRedeemScript);
-
-            txInput.setScriptSig(createBaseInputScriptThatSpendsFromTheFederation(spendingFed, inputRedeemScript));
+            if (!BitcoinUtils.inputHasWitness(pegoutBtcTx, inputIndex)) {
+                txInput.setScriptSig(createBaseInputScriptThatSpendsFromTheFederation(spendingFed, inputRedeemScript));
```
nathanieliov, line 561: Is it possible to have p2wsh and p2sh inputs at the same time? I guess it cannot. If so, you should validate if the pegout has a witness before calling this method and skip calling it.

julianlen: The witness depends on the input, not on the tx.

```
public static boolean inputHasWitness(BtcTransaction btcTx, int inputIndex) {
        TransactionWitness inputWitness = btcTx.getWitness(inputIndex);
        return !inputWitness.equals(TransactionWitness.getEmpty());
```

src/test/java/co/rsk/federate/btcreleaseclient/BtcReleaseClientTest.java
```diff
@@ -488,12 +453,7 @@ void onBestBlock_whenPegoutTxIsCached_shouldNotSignSamePegoutTxAgain() throws Ex
         doReturn(signerPublicKey).when(signer).getPublicKey(BTC.getKeyId());
         doReturn(1).when(signer).getVersionForKeyId(ArgumentMatchers.any(KeyId.class));
         doReturn(ecKey.doSign(new byte[]{})).when(signer).sign(any(KeyId.class), any(SignerMessage.class));
-
-        PowpegNodeSystemProperties powpegNodeSystemProperties = mock(PowpegNodeSystemProperties.class);
-        doReturn(Constants.mainnet()).when(powpegNodeSystemProperties).getNetworkConstants();
         doReturn(true).when(powpegNodeSystemProperties).isPegoutEnabled();
```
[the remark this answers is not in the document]
julianlen: I think it is cleaner to declare it in all the mocks even though true is the default value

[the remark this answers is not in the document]
julianlen: So, should I consider those that don't have any "isPegoutEnable" declared like they have it true?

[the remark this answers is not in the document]
nathanieliov: This `doReturn(true) when (powpegNodeSystemProperties).isPegoutEnabled();` is repeated in a couple of places. What you are trying to do with the new method `getPowpegNodeSystemProperties` is moving the setup of the mock out of the unit test, reducing the boilerplate code. Why not setting `isPegoutEnable` also in that method since it is repeated in a couple of ut?

src/main/java/co/rsk/federate/btcreleaseclient/BtcReleaseClient.java
```diff
@@ -551,13 +551,16 @@ public void onBtcRelease(BtcTransaction signedBtcTx) {
     This way the pegoutBtcTx has the same hash as the one registered in release_requested event topics.
      */
     protected void removeSignaturesFromTransaction(BtcTransaction pegoutBtcTx, Federation spendingFed) {
+        if (pegoutBtcTx.hasWitness()) {
+            return;
```
nathanieliov, line 555: Do you think we should log something when the tx has witness?

nathanieliov (approved):
Well done, left an optional comment.

src/main/java/co/rsk/federate/btcreleaseclient/BtcReleaseClient.java
julianlen: something like

`logger.debug("[removeSignaturesFromTransaction] No signature removed because the signatures are in the witness");`

marcos-iov (approved):

PR #491, opened 2025-09-25
Fix logic in PegUtilsLegacy to correctly check for amount sent in pegin

julia-zack:


src/main/java/co/rsk/federate/bitcoin/BitcoinWrapperImpl.java
```diff
@@ -384,6 +384,21 @@ protected void coinsReceivedOrSent(Transaction tx) {
                 bridgeConstants,
                 activations
             )) {
+
+                boolean amountSentIsBelowMinimum = !PegUtils.allUTXOsToFedAreAboveMinimumPeginValue(
```
nathanieliov, line 388: NIT: The validation is per utxo, not the amount sent. I would rather name this associated with it, e.g: `hasUtxosBelowMinimum`. Wdyt?

nathanieliov (approved):

src/main/java/co/rsk/federate/bitcoin/BitcoinWrapperImpl.java
julia-zack: updated!

marcos-iov (approved):

PR #567, opened 2026-04-14
Fix/use different proofs files for different clients

julia-zack:


src/main/java/co/rsk/federate/io/BtcToRskClientFileStorageInfo.java
```diff
@@ -1,18 +1,15 @@
...
 
-import co.rsk.federate.config.PowpegNodeSystemProperties;
-
 import java.io.File;
 
-public class BtcToRskClientFileStorageInfo implements FileStorageInfo {
-
-    private String pegDirectoryPath;
+public final class BtcToRskClientFileStorageInfo implements FileStorageInfo {
+    private final String pegDirectoryPath;
```
julia-zack: this variable should be called just `directoryPath`, since it belongs to FileStorageInfo interface. I didn't wanna add extra noise; can be done later.

src/main/java/co/rsk/federate/FedNodeRunner.java
```diff
@@ -341,6 +343,21 @@ private void startFederate() throws Exception {
...
+
+    private BtcToRskClientFileStorage createStorageFileForRetiringFedClient() {
+        String fileCustomizer = "retiring";
+        return createStorageFile(fileCustomizer);
+    }
+
+    private BtcToRskClientFileStorage createStorageFile(String fileCustomizer) {
+        FileStorageInfo fileStorageInfo = new BtcToRskClientFileStorageInfo(directoryStorageInfo, fileCustomizer);
+        return new BtcToRskClientFileStorageImpl(fileStorageInfo);
+    }
```
julia-zack: first i thought about creating a `BtcToRskClientFIleStorageFactory` to do this, but felt a bit overkilling.

src/test/java/co/rsk/federate/bitcoin/BitcoinWrapperImplTest.java
```diff
@@ -251,120 +271,79 @@ private static Stream<Arguments> retiringAndActiveFedsArgs() {
...
-            .withFederation(retiringFederation)
-            .build();
-        bitcoinWrapper.addFederationListener(retiringFederation, listener);
+    void coinsReceivedOrSent_peginToRetiringFed_listenerForBothFeds_shouldBeSavedJustInRetiringFedProofsFile(Federation retiringFederation, Federation activeFederation) throws Exception {
+        // arrange
+        setUpActiveFedListener(activeFederation);
+        setUpRetiringFedListener(retiringFederation);
 
-        var peginBtcTxToActiveFed = createPegIn(retiringFederation.getAddress());
+        var peginBtcTxToRetiringFed = createPegIn(retiringFederation.getAddress());
```
julia-zack, line 326: there was a typo here :D

src/test/java/co/rsk/federate/bitcoin/BitcoinWrapperImplTest.java
```diff
@@ -251,120 +271,79 @@ private static Stream<Arguments> retiringAndActiveFedsArgs() {
...
 
-        // Act
-        var peginTxToActiveFed = ThinConverter.toOriginalInstance(originalNetworkParameters.getId(), peginBtcTxToActiveFed);
-        bitcoinWrapper.coinsReceivedOrSent(peginTxToActiveFed);
+        // act
+        var peginTxToRetiringFed = ThinConverter.toOriginalInstance(originalNetworkParameters.getId(), peginBtcTxToRetiringFed);
```
julia-zack, line 329: same here

src/test/java/co/rsk/federate/bitcoin/BitcoinWrapperImplTest.java
```diff
@@ -251,120 +271,79 @@ private static Stream<Arguments> retiringAndActiveFedsArgs() {
 
     @ParameterizedTest
     @MethodSource("retiringAndActiveFedsArgs")
-    void coinsReceivedOrSent_migrationTx_activeFedListener_shouldListenTx(Federation retiringFederation, Federation activeFederation) throws Exception {
```
julia-zack, line 254: replaced this test for when having both listeners, asserting that just the active one saves the proof

src/main/java/co/rsk/federate/io/DirectoryStorageInfo.java
```diff
@@ -0,0 +1,5 @@
+package co.rsk.federate.io;
+
+public interface DirectoryStorageInfo {
```
julianlen, line 3: Do we need this interface?

src/test/java/co/rsk/federate/utils/ClientProofsAssertions.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public static void assertWTxIdIsNotInProofsFile(NetworkParameters networkParameters, BtcToRskClientFileStorage btcToRskClientFileStorage, Transaction tx) throws IOException {
+        BtcToRskClientFileData fileData = btcToRskClientFileStorage.read(networkParameters).getData();
+        Map<Sha256Hash, List<Proof>> transactionProofs = fileData.getTransactionProofs();
+
+        Set<Sha256Hash> txProofsKeySet = transactionProofs.keySet();
+        assertFalse(txProofsKeySet.contains(tx.getWTxId()));
+    }
+
+    public static void assertProofsFileIsEmpty(NetworkParameters networkParameters, BtcToRskClientFileStorage btcToRskClientFileStorage) throws IOException {
```
jeremy-then, line 38: Optional.

Should we have a `assertRetiringFedProofsFileIsEmpty` and `assertActiveFedProofsFileIsEmpty` abstraction to simplify calls from the test and ease reading and let these new methods call this one?
So they are called simply, like the call usually before this one in the tests: `assertWTxIdIsInActiveFedClientProofsFile(migrationTx);`, `assertWTxIdIsInRetiringFedClientProofsFile`.

No big deal really. Just an idea. No need for the change if not useful.

src/test/java/co/rsk/federate/utils/ClientProofsAssertions.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class ClientProofsAssertions {
+
+    public static void assertWTxIdIsInTxsToBeSentMap(BtcToRskClient btcToRskClient, Transaction tx) {
+        assertTrue(btcToRskClient.getTransactionsToSendToRsk().containsKey(tx.getWTxId()));
+    }
+
+    public static void assertWTxIdIsNotInTxsToBeSentMap(BtcToRskClient btcToRskClient, Transaction tx) {
```
julianlen, line 24: This is not being used

src/test/java/co/rsk/federate/BtcToRskClientTest.java
```diff
@@ -1846,19 +1850,55 @@ void setUp() {
...
+            String btcToRskClientFilePrefix = "BtcToRskClient";
+            File directory = new File(directoryStorageInfo.getPath());
+            Wallet wallet = mock(Wallet.class);
+            KitForTests kit = new KitForTests(MAINNET_CONTEXT, directory, btcToRskClientFilePrefix, wallet);
             setUpBitcoinWrapper(kit);
+
+            walletTxs = new HashSet<>();
+            when(wallet.getTransactions(false)).thenReturn(walletTxs);
+
+            int CHAIN_HEIGHT = 4;
```
jeremy-then: Sonar complains here because it seems to be expecting a variable like `int chainHeight = 4;`. Maybe if it was in the global scope of this class, as `final`, it might take it easy.
Up to you.

src/test/java/co/rsk/federate/BtcToRskClientTest.java
```diff
@@ -1932,32 +1964,45 @@ private void setUpClient(Federation federationToListen) throws Exception {
...
                 .withActivationConfig(activationConfig)
                 .build();
+        }
+
+        private void addListener(Federation federationToListen, BtcToRskClient client) {
             bitcoinWrapper.addFederationListener(federationToListen, client);
         }
 
-        private void setUpTx(BtcTransaction btcTx) throws Exception {
+        private void setUpTx(BtcToRskClient client, BtcTransaction btcTx) throws BlockStoreException {
```
julianlen: This exception is not being thrown

jeremy-then (approved):
LGTM.

src/test/java/co/rsk/federate/BtcToRskClientTest.java
```diff
@@ -3349,65 +3394,165 @@ private static Stream<Arguments> activeAndNotActiveFedsArgs() {
...
+         * When two {@link BtcToRskClient} instances shared a single {@link BtcToRskClientFileStorage}
+         * (same on-disk RLP), the retiring client's onBlock could write the file with in-memory state
+         * that did not include txs only the active client had listened to, overwriting
+         * the file and dropping those txs from persistence.
+         * With separate proof files per client, both federations' pending txs must remain stored
+         * after both clients process the same block
+         */
+        @ParameterizedTest
+        @MethodSource("retiringAndActiveFedsArgs")
+        void onBlock_clientForBothFeds_sharedStorage_overwritesProofsFile(Federation retiringFed, Federation activeFed) throws Exception {
```
julianlen: should we keep this test even when in production, there'll be two separated storages?

src/test/java/co/rsk/federate/BtcToRskClientTest.java
```diff
@@ -3349,65 +3394,165 @@ private static Stream<Arguments> activeAndNotActiveFedsArgs() {
...
+            var tx1 = ThinConverter.toOriginalInstance(MAINNET_BTC_PARAMS_STRING, btcTx1);
+            setUpTxConfidence(tx1);
+            listenTx(activeFedClient, tx1);
+            listenTx(retiringFedClient, tx1);
+            // both clients should listen to the same block containing the tx
+            Block blockWithTx1 = buildBlockWithTx(tx1);
+            activeFedClient.onBlock(blockWithTx1);
+            retiringFedClient.onBlock(blockWithTx1);
+
+            // Both clients should have tx1 in their proofs file and their in-memory fileData
```
julianlen, line 3543: wdyt to have a comment explaining that this is to demonstrate the overwriting issue. However, the txs in each proof file and in-memory fileData will be different.

src/test/java/co/rsk/federate/BtcToRskClientTest.java
julianlen: not a huge fan of variables written like this for local variables

src/test/java/co/rsk/federate/utils/ClientProofsAssertions.java
```diff
@@ -0,0 +1,51 @@
...
+
+        Set<Sha256Hash> txProofsKeySet = transactionProofs.keySet();
+        assertTrue(txProofsKeySet.contains(tx.getWTxId()));
+    }
+
+    public static void assertWTxIdIsNotInProofsFile(NetworkParameters networkParameters, BtcToRskClientFileStorage btcToRskClientFileStorage, Transaction tx) throws IOException {
+        BtcToRskClientFileData fileData = btcToRskClientFileStorage.read(networkParameters).getData();
+        Map<Sha256Hash, List<Proof>> transactionProofs = fileData.getTransactionProofs();
+
+        Set<Sha256Hash> txProofsKeySet = transactionProofs.keySet();
```
julianlen: This can be extracted too

julianlen (changes requested):
All looks great, just a few minor comments

src/test/java/co/rsk/federate/utils/ClientProofsAssertions.java
julia-zack: yes, thanks! Already removed here: eacc2ea5dc39927a7344004e41fb9734a7e43ded

src/test/java/co/rsk/federate/utils/ClientProofsAssertions.java
julia-zack: good catch! Done: 1adc638e1a1882e5c4c7975f2726fa4b1d9dc8d2

src/test/java/co/rsk/federate/BtcToRskClientTest.java
julia-zack: thats a good question, i think it does not hurt, over all because of the explanatory comment above

src/test/java/co/rsk/federate/BtcToRskClientTest.java
julia-zack: you right, thanks

Fixed: b9c112378e845d949e2e8e671c7bf2ac6a2399e4

src/test/java/co/rsk/federate/BtcToRskClientTest.java
julia-zack: fixed :)
b9c112378e845d949e2e8e671c7bf2ac6a2399e4

src/main/java/co/rsk/federate/io/DirectoryStorageInfo.java
julia-zack: well, it may seem a bit unnecessary, but we already had the `FIleStorageInfo` interface and `BtcToRskClientFileStorageInfo` impl. So this was the better way to match that

src/test/java/co/rsk/federate/utils/ClientProofsAssertions.java
julia-zack: Good idea, thanks.
Done here: eacc2ea5dc39927a7344004e41fb9734a7e43ded

src/test/java/co/rsk/federate/BtcToRskClientTest.java
julia-zack: wdym? 
I added this javadoc before both tests: https://github.com/rsksmart/powpeg-node/pull/567/changes#diff-34cbd6d3438cd131c3ba5a6f9be3b37210a380e87dfb039fd9d0ac459c0413d9R3444

src/test/java/co/rsk/federate/BtcToRskClientTest.java
jeremy-then: I saw it was already handled in https://github.com/rsksmart/powpeg-node/commit/b9c112378e845d949e2e8e671c7bf2ac6a2399e4. Resolving.

julianlen (approved):

src/main/java/co/rsk/federate/FedNodeRunner.java
marcos-iov: Not a bad idea at all. Although it's not much code it's always a good idea to remove some responsibilities from this already overloaded class

Idea from AI
```
public class BtcToRskClientFileStorageFactory {
       private final DirectoryStorageInfo directoryStorageInfo;
       
       public BtcToRskClientFileStorageFactory(DirectoryStorageInfo directoryStorageInfo) {
           this.directoryStorageInfo = directoryStorageInfo;
       }
       
       public BtcToRskClientFileStorage forActive()  { return create("active"); }
       public BtcToRskClientFileStorage forRetiring(){ return create("retiring"); }
       
       private BtcToRskClientFileStorage create(String customizer) {
           return new BtcToRskClientFileStorageImpl(
               new BtcToRskClientFileStorageInfo(directoryStorageInfo, customizer)
           );
       }
   }
```

src/main/java/co/rsk/federate/io/BtcToRskClientFileStorageInfo.java
marcos-iov: I'd just go for it. It's a private variable, shouldn't affect too much the review

src/test/java/co/rsk/federate/bitcoin/BitcoinWrapperImplTest.java
```diff
@@ -47,35 +47,36 @@ class BitcoinWrapperImplTest {
...
         when(federatorSupport.getConfigForBestBlock()).thenReturn(activations);
 
         // using a temporary directory for testing
-        FileStorageInfo fileStorageInfo = mock(BtcToRskClientFileStorageInfo.class);
-        String pegDir = tempDir.toString();
-        String filePath = tempDir.resolve("btcToRskClient2.rlp").toString();
-        when(fileStorageInfo.getPegDirectoryPath()).thenReturn(pegDir);
-        when(fileStorageInfo.getFilePath()).thenReturn(filePath);
-        btcToRskClientFileStorage = new BtcToRskClientFileStorageImpl(fileStorageInfo);
+        directoryStorageInfo = mock(DirectoryStorageInfo.class);
```
marcos-iov: Any chance we could create the real object instead of the mock?

src/main/java/co/rsk/federate/FedNodeRunner.java
marcos-iov: Could also be used in the tests

julia-zack: what i had done was almost exactly the same yes. Will add it

src/main/java/co/rsk/federate/io/BtcToRskClientFileStorageInfo.java
julia-zack: Updated b7baf576dd65a28bc2a1631712425345d3a06fd5

src/main/java/co/rsk/federate/FedNodeRunner.java
julia-zack: Done in 0904eacbe242114874b902fcac104e69170a30ce

src/test/java/co/rsk/federate/bitcoin/BitcoinWrapperImplTest.java
julia-zack: Something has to be mocked since we are using a tempDir for testing purposes. 
Check if you think this is better: c6d91517baffccbc6dac0e37bd5e6f9d732b9c00
(i think it is)

marcos-iov (approved):

PR #646, opened 2026-07-20
Dont sign when advancing blockchain

julia-zack:


src/main/java/co/rsk/federate/btcreleaseclient/BtcReleaseClient.java
```diff
@@ -515,7 +526,9 @@ protected void signRelease(int signerVersion, ReleaseCreationInformation pegoutC
         } catch (SignerException e) {
             String message = String.format("Error signing pegout created in rsk transaction %s", pegoutCreationRskTxHash);
             logger.error(message, e);
-        } catch (HSMClientException | SignerMessageBuilderException | ReleaseRequirementsEnforcerException e) {
+        } catch (ReleaseRequirementsEnforcerException e) {
+            logger.info("[signRelease] {}", e.getMessage());
```
julia-zack: differentiating RRE from log level (`info` instead of `error`)

